PR #1, opened 2023-02-15
JSON types are declared which are to be used with the siri sx generator

alvak1:


shared-ts/jsonTypes.ts
```diff
@@ -0,0 +1,113 @@
...
+    Version?: Number;
+    Source: Source;
+    Progress: Progress;
+    ValidityPeriod: Period[];
+    PublicationWindow: Period;
+    MiscellaneousReason?: Reason;
+    Planned: boolean;
+    Summary: string;
+    InfoLink?: InfoLink[];
+    Consequences: Consequences[];
```
lj-kpmg, line 96: I'd probably leave consequences out for the time being, we'll be adding them in at a later date

shared-ts/jsonTypes.ts
```diff
@@ -0,0 +1,113 @@
...
+    SituationNumber: string;
+    Version?: Number;
+    Source: Source;
+    Progress: Progress;
+    ValidityPeriod: Period[];
+    PublicationWindow: Period;
+    MiscellaneousReason?: Reason;
+    Planned: boolean;
+    Summary: string;
+    InfoLink?: InfoLink[];
```
lj-kpmg, line 95: I think this should be InfoLinks according to the xsd so the tfn guide might be wrong here. InfoLinks then can contain many InfoLink objects.

shared-ts/jsonTypes.ts
```diff
@@ -0,0 +1,113 @@
...
+    ParticipantRef: string;
+    SituationNumber: string;
+    Version?: Number;
+    Source: Source;
+    Progress: Progress;
+    ValidityPeriod: Period[];
+    PublicationWindow: Period;
+    MiscellaneousReason?: Reason;
+    Planned: boolean;
+    Summary: string;
```
lj-kpmg, line 94: We might need to add Description in as well

shared-ts/jsonTypes.ts
```diff
@@ -0,0 +1,113 @@
...
+    pendingApproval, 
+    approvedDraft, 
+    open, 
+    closing, 
+    closed, 
+    rejected
+}
+
+enum Reason {
+    Accident,
```
lj-kpmg, line 28: There seem to be a few more reasons in the xsd, they also look to start with a lower case letter https://www.siri.org.uk/schema/2.0/xsd/siri_model/siri_situationReasons-v1.1.xsd

shared-ts/jsonTypes.ts
```diff
@@ -0,0 +1,113 @@
...
+    Source: Source;
+    Progress: Progress;
+    ValidityPeriod: Period[];
+    PublicationWindow: Period;
+    MiscellaneousReason?: Reason;
+    Planned: boolean;
+    Summary: string;
+    InfoLink?: InfoLink[];
+    Consequences: Consequences[];
+    Repetitions?: Repetitions[];
```
lj-kpmg, line 97: I think Repetitions isn't an array here but DayType should be, it also needs to come after ValidityPeriod

shared-ts/jsonTypes.ts
```diff
@@ -0,0 +1,113 @@
...
+    References?: Reference;
```
lj-kpmg, line 98: References will need to go after Version since the order matters with the XSD

shared-ts/jsonTypes.ts
alvak1: Removed the above property and its type

shared-ts/jsonTypes.ts
alvak1: Updated the Property name to InfoLinks

shared-ts/jsonTypes.ts
alvak1: Added Description

shared-ts/jsonTypes.ts
alvak1: Added additional reasons and different types of reasons such as environment, personnel and equipment

shared-ts/jsonTypes.ts
alvak1: Updated the Daytype to an array

shared-ts/jsonTypes.ts
alvak1: done

shared-ts/jsonTypes.ts
```diff
@@ -0,0 +1,263 @@
...
+    ParticipantRef: string;
+    SituationNumber: string;
+    Version?: Number;
+    References?: Reference;
+    Source: Source;
+    Progress: Progress;
+    ValidityPeriod: Period[];
+    Repetitions?: Repetitions;
+    PublicationWindow: Period;
+    ReasonType?: MiscellaneousReason | PersonnelReason | EquipmentReason | EnvironmentReason;
```
lj-kpmg, line 244: This would lead to the top level element in the XML being ReasonType, I think what we need is something like:
```
type BasePtSituationElement = {
    CreationTime: Date;
    ParticipantRef: string;
    SituationNumber: string;
    Version?: Number;
    References?: Reference;
    Source: Source;
    Progress: Progress;
    ValidityPeriod: Period[];
    Repetitions?: Repetitions;
    PublicationWindow: Period;
    Planned: boolean;
    Summary: string;
    Description: string;
    InfoLinks?: InfoLink[];
};

type MiscReasonPtSituationElement = BasePtSituationElement & {
  MiscellaneousReason: MiscellaneousReason;
};

type PersonnelReasonPtSituationElement = BasePtSituationElement & {
  PersonnelReason: PersonnelReason;
};

type EquipmentReasonPtSituationElement = BasePtSituationElement & {
  EquipmentReason: EquipmentReason;
};

type EnvironmentReasonPtSituationElement = BasePtSituationElement & {
  EnvironmentReason: EnvironmentReason;
};

export type PtSituationElement = MiscReasonPtSituationElement | PersonnelReasonPtSituationElement | EquipmentReasonPtSituationElement | EnvironmentReasonPtSituationElement;
```

shared-ts/jsonTypes.ts
```diff
@@ -0,0 +1,263 @@
...
+    Planned: boolean;
+    Summary: string;
+    Description: string;
+    InfoLinks?: InfoLink[];
```
lj-kpmg, line 248: This would result in multiple InfoLinks items in the XML when we want a single InfoLinks with multiple InfoLink items below it so we may just need an InfoLinks type:
```
type InfoLinks = {
  InfoLink: InfoLink[];
}
```
then this can be used at the top level `InfoLinks?: InfoLinks;`

lj-kpmg (approved):

PR #3, opened 2023-02-17
feat: Update lambda function to generate SIRI SX

alainajaffer:
## Description

Update lambda function to generate SIRI SX

## Acceptance Criteria

- [ ] Lambda loads json, converts it to valid SIRI-SX and uploads the result to a new S3 bucket (cdd-siri-sx-${stage})

- [ ] Test Disruptions JSON created

- [ ] Write a script which uploads some of the test disruptions json to the S3 bucket to trigger the generator

Example command to put json into s3:

`sh put_file_to_s3.sh put_file_to_s3.sh cdd-siri-sx-aj test_123.json`


shared-ts/jsonTypes.ts
```diff
@@ -106,7 +106,7 @@ export enum MiscellaneousReason {
     undefinedProblem,
     problemsAtBorderPost,
     problemsAtCustomsPost,
-    problemsOnLocalRoad,
+    problemsOnLocalRoad
```
lj-kpmg, line 109: Prettier will want this trailing comma, it's worth installing the prettier and eslint plugins for vscode if you don't have them already

packages/siri-sx-generator/index.ts
```diff
@@ -1,16 +1,47 @@
-import { GetObjectCommand, S3Client } from "@aws-sdk/client-s3";
+import { GetObjectCommand, S3Client, PutObjectCommand } from "@aws-sdk/client-s3";
 import { S3Event } from "aws-lambda";
+import { toXML } from "jstoxml";
 
 const s3Client = new S3Client({ region: "eu-west-2" });
 
+const uploadToS3 = async (data: string, filename: string) => {
+    const s3Client = new S3Client({ region: "eu-west-2" });
```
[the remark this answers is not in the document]
alainajaffer: done

shared-ts/jsonTypes.ts
alainajaffer: done

packages/siri-sx-generator/index.ts
```diff
@@ -1,16 +1,47 @@
...
+
+    // Unique bucket name
+    const bucketName = process.env. SIRI_SX_BUCKET_ARN?.replace("arn:aws:s3:::", "") || "";
```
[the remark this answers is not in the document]
alainajaffer: done

packages/siri-sx-generator/index.ts
```diff
@@ -1,16 +1,45 @@
...
+
+    const xmlData = toXML(JSON.parse(data), config);
+
+    if (!xmlData) {
+        throw Error("Could not generate XML");
+    }
 
-    // eslint-disable-next-line no-console
-    console.log("file contents:", data);
+    await uploadToS3(xmlData, "siri-generated-xml.xml");
```
RobCattonKPMG, line 44: Do we always want to have the same name for the file? Just thinking it will overwrite each time if we do - we could include the timestamp if we wanted different files each time.

packages/siri-sx-generator/index.ts
```diff
@@ -1,16 +1,45 @@
-import { GetObjectCommand, S3Client } from "@aws-sdk/client-s3";
+import { GetObjectCommand, S3Client, PutObjectCommand } from "@aws-sdk/client-s3";
 import { S3Event } from "aws-lambda";
+import { toXML } from "jstoxml";
 
 const s3Client = new S3Client({ region: "eu-west-2" });
 
+const uploadToS3 = async (data: string, filename: string) => {
+    // Unique bucket name
+    const bucketName = process.env.SIRI_SX_BUCKET_NAME || "";
```
RobCattonKPMG, line 9: i wonder if it would be better to cast this to a string instead of defaulting to "" - just because we know it will break either way if its not in the env variables.

packages/siri-sx-generator/index.ts
```diff
@@ -1,16 +1,45 @@
...
+    const keyName = filename;
+
+    const putCommand: PutObjectCommand = new PutObjectCommand({
+        Bucket: bucketName,
+        Key: keyName,
+        Body: data,
+    });
+
+    await s3Client.send(putCommand);
+    console.log("Successfully uploaded data to " + bucketName + "/" + keyName);
```
RobCattonKPMG, line 20:
```suggestion
    console.log(`Successfully uploaded data to ${bucketName}/${keyName}`);
```

packages/siri-sx-generator/index.ts
```diff
@@ -1,16 +1,45 @@
...
+};
+
 export const main = async (event: S3Event): Promise<void> => {
     const bucketName = event.Records[0].s3.bucket.name || "";
     const objectKey = event.Records[0].s3.object.key || "";
 
     const params = { Bucket: bucketName, Key: objectKey };
     const response = await s3Client.send(new GetObjectCommand(params));
-    const data = response.Body?.toString() || "";
+    const data = (await response.Body?.transformToString()) || "";
```
RobCattonKPMG, line 29: How come we're using transformToString() instead of toString()?

packages/siri-sx-generator/index.ts
```diff
@@ -1,16 +1,45 @@
...
+    const config = {
+        indent: "    ",
+    };
+
+    if (!data) {
```
RobCattonKPMG, line 34: If we're doing this, then we dont need to default the data to "" above.

packages/siri-sx-generator/index.ts
```diff
@@ -1,16 +1,45 @@
...
+        throw Error("No data found");
+    }
+
+    const xmlData = toXML(JSON.parse(data), config);
+
+    if (!xmlData) {
```
RobCattonKPMG, line 40: Does this ever happen? I would have thought that the 'toXML' function throws an error if it cant convert the data to XML, instead of setting it to undefined, but I dont know how the library works.

packages/siri-sx-generator/test/siriSXPublishToS3.ts
```diff
@@ -0,0 +1,41 @@
+import { SituationExchangeDelivery, SourceType, Progress, MiscellaneousReason } from "../../../shared-ts/jsonTypes";
+
+export const jsonData: SituationExchangeDelivery = {
+    ResponseTimestamp: "2021-05-11T08:51:07.016Z",
+    Situations: {
+        PtSituationElement: [
+            {
+                CreationTime: "2020-05-01T08:45:58Z",
+                ParticipantRef: "ItoWorld",
```
RobCattonKPMG, line 9: What is this file used for?

shared-ts/jsonTypes.ts
```diff
@@ -212,16 +212,21 @@ export type Source = {
 };
 
 export type Period = {
```
lj-kpmg, line 214: Just noticed the filename is being changed back here, I'd leave it as index.d.ts since it's the more standard way to declare a type file

packages/siri-sx-generator/index.ts
lj-kpmg: @RobCattonKPMG That's the function exposed to convert the stream returned by the AWS SDK into a string, `toString` won't work here

RobCattonKPMG: Fair enough! Haven't seen it before

packages/siri-sx-generator/index.ts
lj-kpmg: @RobCattonKPMG I wouldn't worry about this too much for now, this ticket is just proving out the conversion, we'll have to think about the best way to store the files in S3

packages/siri-sx-generator/package.json
```diff
@@ -8,7 +8,9 @@
     },
     "devDependencies": {
         "@types/aws-lambda": "^8.10.110",
+        "@types/jstoxml": "^2.0.2",
         "@types/node": "^18.13.0",
+        "jstoxml": "^3.2.6",
```
lj-kpmg, line 13: This would be better off as a dependency rather than a devDependency

packages/siri-sx-generator/index.ts
RobCattonKPMG: No worries

packages/siri-sx-generator/index.ts
alainajaffer: done

packages/siri-sx-generator/index.ts
alainajaffer: couldn't find it in docs so this is precautionary

shared-ts/jsonTypes.ts
alainajaffer: done

packages/siri-sx-generator/package.json
alainajaffer: done

packages/siri-sx-generator/test/siriSXPublishToS3.ts
alainajaffer: used for the ts script

packages/siri-sx-generator/index.ts
alainajaffer: done

packages/siri-sx-generator/index.ts
alainajaffer: refactored this anyways

packages/siri-sx-generator/test/siriSXPublishToS3.ts
RobCattonKPMG: which script is this sorry?

alainajaffer: the bash script to put json into s3

packages/siri-sx-generator/index.ts
alainajaffer: export function toXML(obj?: XmlElement | XmlElement[], options?: XmlOptions): string;

RobCattonKPMG: It seems like it would either return a string or error, but it cant hurt having this check in. No worries.

lj-kpmg (approved):

PR #7, opened 2023-02-28
feat: create footer component

RobCattonKPMG:
Editing the footer component to match other gov.uk pages.


site/tailwind.config.js
```diff
@@ -22,6 +22,9 @@ module.exports = {
                     .filter((i) => i)
                     .reduce((acc, i) => ({ ...acc, [i]: `${i / (baseFontSize / 4)}rem` }), {}),
             }),
+            backgroundImage: {
+                "govuk-crest": "url('../public/assets/images/govuk-crest-2x.png')",
```
lj-kpmg, line 26: We can use the absolute path here `/assets/images/govuk-crest-2x.png` since the relative path might not work when the site is deployed

lj-kpmg (approved):

PR #9, opened 2023-03-01
Added fields for create disruption page

alvak1:


site/components/TimeSelector.tsx
```diff
@@ -0,0 +1,53 @@
...
+                    <div className="govuk-form-group">
+                        <label className="govuk-label govuk-date-input__label" htmlFor={`${startOrEnd}-hour-input`}>
+                            Hours
+                        </label>
+                        <input
+                            className={`govuk-input govuk-date-input__input govuk-input--width-2 ${
+                                errors.length > 0 ? 'govuk-input--error' : ''
+                            } `}
+                            id={`${startOrEnd}-hour-input`}
+                            name={`${startOrEnd}Hours`}
```
RobCattonKPMG, line 27: the below input has a name of name={`${startOrEnd}Minute`} , but this one has 'Hours' - it would make sense to have them both be plural or neither plural, i.e. Minutes and Hours, or Minute and Hour

site/pages/create-disruption.tsx
```diff
@@ -1,5 +1,211 @@
+import DateSelector from "components/DateSelector";
+import TimeSelector from "components/TimeSelector";
+import { BaseLayout } from "components/layout/Layout";
+
+const title = "Create Disruptions";
+const description = "Create Disruptions page for the Create Transport Disruptions Service";
+
+enum Reason {
```
RobCattonKPMG, line 8: if these are all mapping to the exact same value as a string, would it make sense for this to just be a string array? And then lower down, instead of using Object.keys , we can just map over the array? Im not sure what value we get from using an enum here.

RobCattonKPMG: Although in saying that, maybe the benefit is because we shouldnt be using camelCase values as what the user selects - should it be `roadWorks = "Road works",`

site/pages/create-disruption.tsx
```diff
@@ -1,5 +1,211 @@
...
+    roadWorks = "roadWorks",
+    vandalism = "vandalism",
+    routeDiversion = "routeDiversion",
+    specialEvent = "specialEvent",
+}
+
+const renderOption = (text: string) => {
```
RobCattonKPMG, line 15: If we do the above, then we dont need to use this function

site/pages/create-disruption.tsx
```diff
@@ -1,5 +1,211 @@
...
+                                name="associatedLink"
+                                type="text"
+                            />
+                        </div>
+                        <div className="govuk-form-group">
+                            <label className="govuk-label" htmlFor="summary">
+                                Reason for disruption
+                            </label>
+                            <select className="govuk-select w-3/4" id="disruption-reason" name="disruptionReason">
+                                {Object.keys(Reason).sort().map(renderOption)}
```
RobCattonKPMG, line 90: will the select options not need a value also? i.e. the value would be the key of the enum

site/pages/create-disruption.tsx
```diff
@@ -1,5 +1,211 @@
...
+                            inputs={{
+                                dayInput: "",
+                                monthInput: "",
+                                yearInput: "",
+                            }}
+                        />
+
+                        <legend className="govuk-fieldset__legend govuk-!-padding-top-4">What is the end time?</legend>
+                        <div id="end-time-hint" className="govuk-hint">
+                            Enter in format xxxx
```
RobCattonKPMG, line 147: Is this a helpful hint? If I'm told to enter in format xxxx, it seems a little unhelpful. HH:MM and an example would be good I think

site/components/TimeSelector.tsx
alainajaffer: done on my branch @alvak1  you can pull those changes

site/pages/create-disruption.tsx
alainajaffer: discussing with the design team and will update it appropriately

alainajaffer: resolved to be HH:MM @alvak1  changes are on my branch you can pull them

site/pages/create-disruption.tsx
lj-kpmg: I'd maybe approach it slightly differently and use the types defined in the shared-ts folder here, something like:
```
const reasons: {
    value:
        | siriTypes.MiscellaneousReason
        | siriTypes.PersonnelReason
        | siriTypes.EnvironmentReason
        | siriTypes.EquipmentReason;
    display: string;
}[] = [
    {
        value: siriTypes.MiscellaneousReason.accident,
        display: "Accident",
    },
    {
        value: siriTypes.PersonnelReason.staffAbsence,
        display: "Staff absence",
    },
];
```

alvak1: Updated the key value pairs and code to access key and value. The enum was used because the XML element for Reason is also an enum and we could reuse the shared type when we have all the values

site/pages/create-disruption.tsx
alvak1: Removed this method as the enum values have user friendly names

site/pages/create-disruption.tsx
alvak1: Value added for options

site/pages/create-disruption.tsx
```diff
@@ -1,24 +1,104 @@
...
+
 const CreateDisruption = ({ inputs }: CreateDisruptionProps): ReactElement => {
     return (
         <BaseLayout title={title} description={description}>
             <>
                 <div className="govuk-form-group">
-                    <h1 className="govuk-heading-l">
-                        When is the disruption?
-                    </h1>
+                    <h1 className="govuk-heading-xl">Create a new Distruption</h1>
```
lj-kpmg, line 36: Typo Distruption -> Disruption

site/pages/create-disruption.tsx
```diff
@@ -1,24 +1,104 @@
...
+                    <fieldset className="govuk-fieldset">
```
alainajaffer, line 37: I assume we should have a fieldset for each sub section

site/pages/create-disruption.tsx
```diff
@@ -1,24 +1,104 @@
...
+                        <div className="govuk-form-group">
+                            <label className="govuk-label mb-3" htmlFor="disruption-type">
+                                Type of disruption
```
alainajaffer, line 40: should be bold use  <h3 className="govuk-heading-s govuk-!-margin-bottom-0">

site/pages/create-disruption.tsx
```diff
@@ -1,24 +1,108 @@
...
         <BaseLayout title={title} description={description}>
             <>
                 <div className="govuk-form-group">
-                    <h1 className="govuk-heading-l">
-                        When is the disruption?
-                    </h1>
+                    <h1 className="govuk-heading-xl">Create a new Distruption</h1>
+                    <fieldset className="govuk-fieldset">
+                        <div className="govuk-form-group">
+                            <label className="govuk-label mb-3" htmlFor="disruption-type">
```
lj-kpmg, line 43: This should be a legend rather than a label and should be outside the form group https://design-system.service.gov.uk/components/fieldset/

site/pages/create-disruption.tsx
lj-kpmg: I think an array makes more sense here, the enums in the shared types won't be changing so we can still make use of those now to give us some type safety for the array, when Steve gives us the actual values we can just update the list

site/pages/create-disruption.tsx
lj-kpmg: Any related inputs should be in a fieldset, so for this section only the planned and unplanned fields really need to be in one

site/pages/create-disruption.tsx
```diff
@@ -10,58 +11,195 @@ interface CreateDisruptionProps {
...
 }
 
+enum Reason {
+    roadWorks = "Road Works",
+    vandalism = "Vandalism",
+    routeDiversion = "Route Diversion",
+    specialEvent = "Special Event",
+}
+
+const reasons: {
```
RobCattonKPMG, line 21: Should we move these into the constants file?

site/pages/create-disruption.tsx
```diff
@@ -10,58 +11,195 @@ interface CreateDisruptionProps {
...
+        reason: "Route Diversion",
+    },
+    {
+        value: MiscellaneousReason.specialEvent,
+        reason: "Special Event",
+    },
+];
+
+const getReasonOptions = (): JSX.Element[] => {
+    let options: JSX.Element[] = [];
```
RobCattonKPMG, line 44: I think this can be a const - it only needs to be a let if we completely reassign the array.

site/pages/create-disruption.tsx
alvak1: done

site/pages/create-disruption.tsx
alvak1: done

site/pages/create-disruption.tsx
alvak1: done, all labels updated to match above styles

site/pages/create-disruption.tsx
alvak1: done

site/pages/create-disruption.tsx
alvak1: the variable is now removed and not required

site/pages/create-disruption.tsx
```diff
@@ -1,7 +1,8 @@
-import { ReactElement } from "react";
-import { DisruptionsDateTimeInfo } from "../components/DisruptionDateTimeInfo";
+import React, { ReactElement } from "react";
+import { DisruptionInfo } from "interfaces";
+import { DisruptionsDateTimeInfo } from "components/DisruptionDateTimeInfo";
+import { MiscellaneousReason, PersonnelReason, EnvironmentReason, EquipmentReason } from "../../shared-ts/siriTypes";
```
lj-kpmg, line 4: These can be imported from the workspace package rather than using the relative path: `@create-disruptions-data/shared-ts/siriTypes`, then we shouldn't need to enable the externalDir feature

site/pages/create-disruption.tsx
```diff
@@ -10,58 +11,195 @@ interface CreateDisruptionProps {
...
+    {
+        value: MiscellaneousReason.specialEvent,
+        reason: "Special Event",
+    },
+];
+
+const getReasonOptions = (): JSX.Element[] => {
+    let options: JSX.Element[] = [];
+
+    options.push(<option value="" key=""></option>);
```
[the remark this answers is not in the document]
alvak1: updated to match default value in figma

site/pages/create-disruption.tsx
alvak1: done

site/pages/create-disruption.tsx
RobCattonKPMG: Unsure what you mean @alvak1  - what variable?

site/pages/create-disruption.tsx
```diff
@@ -10,58 +11,192 @@ interface CreateDisruptionProps {
...
+                                    name="description"
+                                    rows={5}
                                 />
-                                <label className="govuk-label govuk-radios__label" htmlFor="disruption-repeats">
-                                    Yes
-                                </label>
                             </div>
-                            <div className="govuk-radios__item">
+                            <div className="govuk-form-group">
+                                <legend className="govuk-fieldset__legend govuk-!-padding-top-2">
```
lj-kpmg, line 122: This should be a label rather than a legend

site/pages/create-disruption.tsx
```diff
@@ -10,58 +11,192 @@ interface CreateDisruptionProps {
...
+                                    id="associated-link"
+                                    name="associatedLink"
+                                    type="text"
                                 />
-                                <label className="govuk-label govuk-radios__label" htmlFor="disruption-does-not-repeat">
-                                    No
-                                </label>
+                            </div>
+                            <div className="govuk-form-group">
+                                <legend className="govuk-fieldset__legend govuk-!-padding-top-2">
```
lj-kpmg, line 135: This should be a label rather than a legend

site/pages/create-disruption.tsx
```diff
@@ -10,58 +11,192 @@ interface CreateDisruptionProps {
...
                         </div>
-                    </fieldset>
+                        <div className="govuk-form-group">
+                            <h1 className="govuk-heading-l">When is the disruption?</h1>
 
-                    <h1 className="govuk-heading-l govuk-!-padding-top-8">
-                        When does the disruption need to be published?
-                    </h1>
+                            <DisruptionsDateTimeInfo inputs={inputs} isDisruptionValidity />
+                            <legend
```
lj-kpmg, line 149: This legend and the below radio buttons should be wrapped in a fieldset

site/pages/create-disruption.tsx
```diff
@@ -1,67 +1,182 @@
...
 }
 
+const getReasonOptions = (): JSX.Element[] => {
+    const options: JSX.Element[] = [];
+
+    options.push(
+        <option value="" key="">
+            Select Reason
+        </option>,
+    );
```
RobCattonKPMG, line 23:
```suggestion
    const options: JSX.Element[] = [ <option value="" key="">
            Select Reason
        </option>];
```

We can just start the array with the value inside it

site/pages/create-disruption.tsx
alvak1: Apologies, mistook this variable to the another variable that was deleted. I have moved it to the constants file

site/pages/create-disruption.tsx
```diff
@@ -1,142 +1,227 @@
...
+const title = "Create Disruptions";
+const description = "Create Disruptions page for the Create Transport Disruptions Service";
 
 interface CreateDisruptionProps {
-    inputs: DisruptionValidity
+    inputs: DisruptionInfo;
+    errors: ErrorInfo[];
+}
+
+enum Reason {
```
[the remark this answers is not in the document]
alvak1: done

site/pages/create-disruption.tsx
```diff
@@ -1,142 +1,227 @@
...
-import { BaseLayout } from "components/layout/Layout"
-import { DisruptionValidity } from "interfaces";
-
-const title = 'Create Disruptions';
-const description = 'Create Disruptions page for the Create Transport Disruptions Service';
+import { BaseLayout } from "components/layout/Layout";
+import { DisruptionInfo, ErrorInfo } from "interfaces";
+import { DisruptionsDateTimeInfo } from "components/DisruptionDateTimeInfo";
+import ErrorSummary from "../components/ErrorSummary";
+import { MiscellaneousReason, PersonnelReason, EnvironmentReason, EquipmentReason } from "../../shared-ts/jsonTypes";
```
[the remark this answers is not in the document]
alvak1: done

site/pages/create-disruption.tsx
alvak1: duplicate comment

site/pages/create-disruption.tsx
```diff
@@ -10,58 +11,188 @@ interface CreateDisruptionProps {
...
+
 const CreateDisruption = ({ inputs }: CreateDisruptionProps): ReactElement => {
     return (
         <BaseLayout title={title} description={description}>
-            <>
-                <div className="govuk-form-group">
-                    <h1 className="govuk-heading-l">When is the disruption?</h1>
-
-                    <DisruptionsDateTimeInfo inputs={inputs} isDisruptionValidity />
+            <fieldset className="govuk-fieldset">
```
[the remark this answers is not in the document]
alvak1: done

site/pages/create-disruption.tsx
```diff
@@ -10,58 +11,192 @@ interface CreateDisruptionProps {
...
-                                    name="disruptionRepeats"
-                                    type="radio"
-                                    value="yes"
+                            <div className="govuk-form-group">
+                                <legend className="govuk-fieldset__legend govuk-!-padding-top-2">
+                                    <h3 className="govuk-heading-s govuk-!-margin-bottom-0" id="disruption-type">
+                                        Type of disruption
+                                    </h3>
+                                </legend>
+                                <fieldset className="govuk-fieldset">
```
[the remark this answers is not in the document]
alvak1: done

site/pages/create-disruption.tsx
```diff
@@ -10,58 +11,192 @@ interface CreateDisruptionProps {
...
+                                            className="govuk-label govuk-radios__label"
+                                            htmlFor="disruption-unplanned"
+                                        >
+                                            Unplanned
+                                        </label>
+                                    </div>
+                                </fieldset>
+                            </div>
+                            <div className="govuk-form-group">
+                                <legend className="govuk-fieldset__legend govuk-!-padding-top-2">
```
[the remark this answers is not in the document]
alvak1: done

site/pages/create-disruption.tsx
```diff
@@ -10,58 +11,192 @@ interface CreateDisruptionProps {
...
+                                    <h3 className="govuk-heading-s govuk-!-margin-bottom-0" id="summary">
+                                        Summary
+                                    </h3>
+                                </legend>
+                                <input className="govuk-input w-3/4" id="summary" name="summary" type="text" />
+                            </div>
+                            <div className="govuk-form-group">
+                                <legend className="govuk-fieldset__legend govuk-!-padding-top-2">
```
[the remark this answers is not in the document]
alvak1: done

site/pages/create-disruption.tsx
alvak1: done

site/pages/create-disruption.tsx
alvak1: done

site/pages/create-disruption.tsx
alvak1: done

site/pages/create-disruption.tsx
alvak1: done

lj-kpmg (approved):

PR #11, opened 2023-03-01
feat: Create disruption publication

alainajaffer:
Final segment of the create disruptions page

- Requires snapshot tests depending on when @alvak1's pr goes in first or not


site/package.json
```diff
@@ -31,6 +31,7 @@
         "@babel/preset-react": "^7.18.6",
         "@babel/preset-typescript": "^7.21.0",
         "@testing-library/react": "^14.0.0",
+        "@types/lodash": "^4.14.191",
```
lj-kpmg, line 34: We should have lodash itself as well in dependencies

alainajaffer: done

lj-kpmg (approved):

PR #13, opened 2023-03-02
feat: date picker component

RobCattonKPMG:


site/components/DateSelector.tsx
```diff
@@ -1,81 +1,124 @@
...
 
+const renderWeekPickerDay = (
+    _date: Date,
+    _selectedDates: Array<Date | null>,
+    pickersDayProps: PickersDayProps<Date>,
+) => {
+    return (
+        <PickersDay
+            {...pickersDayProps}
+            sx={{
```
lj-kpmg, line 60: Looks like material ui lets you add classes to the components as well so we could use tailwind directly here with something like:
```
classes={{
    selected: "!bg-govBlue",
    dayWithMargin:
        "focus:!border focus:!border-solid hover:!border hover:!border-solid hover:!border-govBlue focus:!border-govYellow",
    }}
```

site/components/DateSelector.tsx
```diff
@@ -1,81 +1,124 @@
...
+const inputBox = (
+    startOrEnd: "start" | "end",
+    hasErrors: boolean,
+    inputRef: React.Ref<any> | undefined,
+    inputProps: InputBaseComponentProps | undefined,
+    InputProps: Partial<FilledInputProps> | Partial<OutlinedInputProps> | undefined,
+    disabled: boolean,
+    isPublishInfo: boolean
+) => (
+    <div className="govuk-date-input flex items-center" id={`${startOrEnd}-date`}>
```
lj-kpmg, line 32: We could also add a tailwind class in here to change the calendar color: `[&_.MuiSvgIcon-root]:fill-govBlue`, that way we shouldn't need to use the custom mui theme

site/package.json
```diff
@@ -13,10 +13,15 @@
     },
     "dependencies": {
         "@create-disruptions-data/shared-ts": "file:../shared-ts",
+        "@emotion/react": "^11.10.6",
```
lj-kpmg, line 16: Are these emotion libraries needed? Seems to work ok for me without them

site/components/DateSelector.tsx
```diff
@@ -1,81 +1,124 @@
...
+        <div className="govuk-date-input__item govuk-!-margin-right-0">
+            <div className="govuk-form-group">
+                <input
+                    className={`govuk-input govuk-date-input__input govuk-input--width-6 ${
+                        hasErrors ? "govuk-input--error" : ""
+                    } `}
+                    id={`${isPublishInfo && 'publish-'}${startOrEnd}-day-input`}
+                    name={`${startOrEnd}DateDay`}
```
alvak1, line 40: The name attribute value should contain a prefix of validity/publish

site/components/DateSelector.tsx
```diff
@@ -1,81 +1,124 @@
...
+    inputRef: React.Ref<any> | undefined,
+    inputProps: InputBaseComponentProps | undefined,
+    InputProps: Partial<FilledInputProps> | Partial<OutlinedInputProps> | undefined,
+    disabled: boolean,
+    isPublishInfo: boolean
+) => (
+    <div className="govuk-date-input flex items-center" id={`${startOrEnd}-date`}>
+        <div className="govuk-date-input__item govuk-!-margin-right-0">
+            <div className="govuk-form-group">
+                <input
```
lj-kpmg, line 35: Since we only have a single input, the wrapping fieldset in `DisruptionDateTimeInfo.tsx` should just be a form group with a label inside instead

site/components/DisruptionDateTimeInfo.tsx
```diff
@@ -29,12 +28,9 @@ export const DisruptionsDateTimeInfo = ({
...
                 startOrEnd="start"
-                inputs={{
-                    dayInput: isDisruptionValidity ? inputs.validityStartDateDay : inputs.publishStartDateDay,
-                    monthInput: isDisruptionValidity ? inputs.validityStartDateMonth : inputs.publishStartDateMonth,
-                    yearInput: isDisruptionValidity ? inputs.validityStartDateYear : inputs.publishStartDateYear,
-                }}
-                type={isDisruptionValidity ? "validity" : "publish"}
+                input={!!inputs.validityStartDate ? new Date(inputs.validityStartDate) : null}
+                disabled={false}
+                isPublishInfo={false}
```
lj-kpmg, line 33: This needs to be defined by `isDisruptionValidity`

site/components/DisruptionDateTimeInfo.tsx
```diff
@@ -72,12 +68,9 @@ export const DisruptionsDateTimeInfo = ({
...
                 startOrEnd="end"
-                inputs={{
-                    dayInput: isDisruptionValidity ? inputs.validityEndDateDay : inputs.publishEndDateDay,
-                    monthInput: isDisruptionValidity ? inputs.validityEndDateMonth : inputs.publishEndDateMonth,
-                    yearInput: isDisruptionValidity ? inputs.validityEndDateYear : inputs.publishEndDateYear,
-                }}
-                type={isDisruptionValidity ? "validity" : "publish"}
+                input={!!inputs.validityEndDate ? new Date(inputs.validityEndDate) : null}
+                disabled={false}
+                isPublishInfo
```
lj-kpmg, line 73: Same here

site/components/DateSelector.tsx
```diff
@@ -1,81 +1,124 @@
...
+                    className={`govuk-input govuk-date-input__input govuk-input--width-6 ${
+                        hasErrors ? "govuk-input--error" : ""
+                    } `}
+                    id={`${isPublishInfo && 'publish-'}${startOrEnd}-day-input`}
```
lj-kpmg, line 39: This will render as "falsestart-day-input" when `isPublishInfo` is false, a ternary will work

site/package.json
RobCattonKPMG: Yeah it says in the docs they're needed. It works without them... for a while, and every now and again it errors (when running locally). I initially didnt have them, but it error'd every now and again.

site/components/DateSelector.tsx
RobCattonKPMG: added

site/components/DateSelector.tsx
RobCattonKPMG: added

site/components/DateSelector.tsx
RobCattonKPMG: added

site/components/DisruptionDateTimeInfo.tsx
RobCattonKPMG: added

site/components/DisruptionDateTimeInfo.tsx
RobCattonKPMG: added

site/components/DateSelector.tsx
RobCattonKPMG: changed

site/components/DateSelector.tsx
lj-kpmg: So in `DisruptionDateTimeInfo.tsx`we have the following for the start date and end date:
```
<fieldset
    className="govuk-fieldset"
    role="group"
    aria-describedby={`${isDisruptionValidity ? "validity" : "publish"}-start-date-hint`}
>
    <legend className="govuk-fieldset__legend govuk-!-padding-top-2">
        <h3 className="govuk-heading-s govuk-!-margin-bottom-0">What is the start date?</h3>
    </legend>
    <div id={`${isDisruptionValidity ? "validity" : "publish"}-start-date-hint`} className="govuk-hint">
        Enter in format DD/MM/YYYY
    </div>
    <DateSelector
        errors={[]}
        startOrEnd="start"
        input={!!inputs.validityStartDate ? new Date(inputs.validityStartDate) : null}
        disabled={false}
        isPublishInfo={!isDisruptionValidity}
    />
</fieldset>
```

but now since we only have a single field this can be changed to just be a label rather than a fieldset and legend to something like:

```
    <label
        className="govuk-label govuk-label--s"
        htmlFor={`${isDisruptionValidity ? "validity" : "publish"}-start-date`}
    >
        What is the start date?
    </label>
    <div id={`${isDisruptionValidity ? "validity" : "publish"}-start-date-hint`} className="govuk-hint">
        Enter in format DD/MM/YYYY
    </div>
    <DateSelector
        errors={[]}
        startOrEnd="start"
        input={!!inputs.validityStartDate ? new Date(inputs.validityStartDate) : null}
        disabled={false}
        isPublishInfo={!isDisruptionValidity}
    />
</div>
```

we just need to make sure the value passed into `htmlFor` on the label matches the id of the input and that we also pass an `aria-describedby` to the input that matches the hint id

RobCattonKPMG: There is a form-group inside the DateSelector - does that need removing? Or is it normal to have a form group inside a form group?

lj-kpmg: It might actually be nicer if rather than passing through `isPublishInfo`, we could just pass through the id we want to apply to the input, then the DateSelector doesn't need to bother checking if it's publish or not

lj-kpmg: > There is a form-group inside the DateSelector - does that need removing? Or is it normal to have a form group inside a form group?

It's fine to nest them yeah

RobCattonKPMG: ok cool! will make both changes.

site/components/DateSelector.tsx
```diff
@@ -1,81 +1,100 @@
...
+const inputBox = (
+    startOrEnd: "start" | "end",
+    hasErrors: boolean,
+    inputRef: React.Ref<any> | undefined,
+    inputProps: InputBaseComponentProps | undefined,
+    InputProps: Partial<FilledInputProps> | Partial<OutlinedInputProps> | undefined,
+    disabled: boolean,
+    inputId: string,
+) => (
+    <div className="govuk-date-input flex items-center [&_.MuiSvgIcon-root]:fill-govBlue" id={`${startOrEnd}-date`}>
```
lj-kpmg, line 28: Is this id being used for anything? We might be able to remove startOrEnd as a prop if not

site/components/DateSelector.tsx
```diff
@@ -1,81 +1,100 @@
...
+        <div className="govuk-date-input__item govuk-!-margin-right-0">
+            <div className="govuk-form-group">
+                <input
+                    className={`govuk-input govuk-date-input__input govuk-input--width-6 ${
+                        hasErrors ? "govuk-input--error" : ""
+                    } `}
+                    id={inputId}
+                    name={`${inputId.includes("publish") ? "publish" : "validity"}-${startOrEnd}DateDay`}
```
lj-kpmg, line 36: I'd be tempted to pass this through as a prop as well, then this component doesn't need to care about publish or validity and can leave it up to the parent, also I'd remove the hyphen from the name since I think we generally use camelcase for the input names

site/components/DateSelector.tsx
```diff
@@ -1,81 +1,100 @@
...
+    input = null,
+    startOrEnd,
+    errors = [],
+    disabled,
+    inputId,
+}: DateSelectorProps): ReactElement => {
+    const [value, setValue] = useState(input);
+
+    return (
+        <FormElementWrapper errors={errors} errorId={`${startOrEnd}-day-input`} errorClass="govuk-date-input--error">
```
lj-kpmg, line 73: I think FormElementWrapper should wrap the input directly and the errorId should be the same as the inputs, so we can just use the `inputId` prop

site/components/DateSelector.tsx
```diff
@@ -1,81 +1,100 @@
...
+                            startOrEnd,
+                            errors.length > 0,
+                            inputRef,
+                            inputProps,
+                            InputProps,
+                            disabled,
+                            inputId,
+                        );
+                    }}
+                    disablePast={startOrEnd === "end"}
```
lj-kpmg, line 92: If we pass `disablePast` as a prop then we can get rid of `startOrEnd`

site/components/DateSelector.tsx
RobCattonKPMG: removed

site/components/DateSelector.tsx
RobCattonKPMG: Yeah sure

site/components/DateSelector.tsx
RobCattonKPMG: done

lj-kpmg (approved):

PR #16, opened 2023-03-06
feat: Added client side error handling

RobCattonKPMG:


site/components/DateSelector.test.tsx
```diff
@@ -58,6 +92,11 @@ describe("DateSelector", () => {
                     inputId={"publish-end-date"}
                     inputName={"publishStartDateDay"}
                     disabled
+                    pageState={blankInputs}
+                    // eslint-disable-next-line @typescript-eslint/no-empty-function
+                    updatePageState={() => {}}
+                    // eslint-disable-next-line @typescript-eslint/no-empty-function
+                    updaterFunction={() => {}}
```
alainajaffer, line 99: would a file disable of this rule be more appropriate?

RobCattonKPMG: I dont really have much of an opinion - @lj-kpmg ?

lj-kpmg: Yeah, put it as a file level disable if it's being used in a few places

site/components/DateSelector.tsx
```diff
@@ -15,37 +16,62 @@ interface DateSelectorProps {
...
+                    name={inputName}
+                    type="text"
+                    ref={inputRef}
+                    {...inputProps}
+                    disabled={disabled}
+                    placeholder={disabled ? "N/A" : "DD/MM/YYYY"}
+                    onBlur={(e) => {
+                        const input = e.target.value;
+                        if (!input) {
+                            updaterFunction(pageState, updatePageState, inputId as keyof PageInputs, input, {
```
lj-kpmg, line 62: If inputId needs to be a key of PageInputs can we not just enforce that in the props rather than using `as` here?

site/pages/create-disruption.test.tsx
```diff
@@ -1,28 +1,45 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import React from "react";
```
lj-kpmg, line 2: React doesn't need to be imported

RobCattonKPMG: Ill try removing it but it was making the tests fail when it wasnt there.

site/components/DateSelector.tsx
RobCattonKPMG: Yep, changed.

site/pages/create-disruption.test.tsx
RobCattonKPMG: Ah it must have been failing because of something else - removed.

site/components/DateSelector.test.tsx
RobCattonKPMG: disabled it for the file

lj-kpmg (approved):

PR #21, opened 2023-03-08
feat:  Time and Date fields should be cleared and disabled if "No end date/time" checked

alainajaffer:
On the create disruptions form you should be able to select a date and time then click on the no date/time checkbox. This should cause the input to clear and become disabled. The state should also be cleared.


alainajaffer:
closing due to large refactor on main - to be redone

PR #25, opened 2023-03-10
feat: Consequence operator page

alainajaffer:
As per wireframes this page follows after the add a consequence page - type of consequence.

This ticket is client side error handling and UI only


site/constants/index.ts
```diff
@@ -33,6 +34,42 @@ export const DISRUPTION_REASONS: {
     },
 ];
 
+export const DISRUPTION_SEVERITIES: {
+    value: Severity;
+    display: string;
+}[] = [
+    { value: Severity.unknown, display: "Unknown" },
+    { value: Severity.verySlight, display: "Very Slight" },
```
RobCattonKPMG, line 42: I think we should use 'Very slight' (sentence case is the default for non-name values).

site/constants/index.ts
```diff
@@ -33,6 +34,42 @@ export const DISRUPTION_REASONS: {
...
+    },
+    {
+        value: Severity.normal,
+        display: "Normal",
+    },
+    {
+        value: Severity.severe,
+        display: "Severe",
+    },
+    { value: Severity.verySevere, display: "Very Severe" },
```
RobCattonKPMG, line 55: As per above comment

site/pages/create-consequence-operator.tsx
```diff
@@ -0,0 +1,209 @@
...
+    return (
+        <BaseLayout title={title} description={description}>
+            <form action="/api/createConsequenceOperator" method="post">
+                <>
+                    <div className="govuk-form-group">
+                        <h1 className="govuk-heading-xl">Add a Consequence</h1>
+                        <Table
+                            rows={[
+                                {
+                                    header: "Mode of Transport",
```
RobCattonKPMG, line 67: Mode of transport

site/pages/create-consequence-operator.tsx
```diff
@@ -0,0 +1,209 @@
...
+                            ]}
+                            inputName="remove-from-journey-planners"
+                            stateUpdater={stateUpdater}
+                            value={pageState.inputs["remove-from-journey-planners"]}
+                        />
+
+                        <TimeSelector<PageInputs>
+                            display="How long is the disruption delay?"
+                            displaySize="l"
+                            hint="Enter the time in 24hr format. For example 0900 is 9am, 1730 is 5:30pm"
```
RobCattonKPMG, line 135: The hint text needs to be slightly different here, as they're providing an amount of time, not a time. Im actually curious if we need a different input here, because they could enter 4800 (for 48 hours) and that would be acceptable. Or 0090 for 90 minutes.

site/pages/create-consequence-operator.tsx
```diff
@@ -0,0 +1,209 @@
+import Link from "next/link";
```
lj-kpmg, line 1: This file will just need renaming to `create-consequence-operator.page.tsx`

site/pages/create-consequence-operator.tsx
alainajaffer: done renamed hint

site/pages/create-consequence-operator.tsx
alainajaffer: done

site/pages/create-consequence-operator.tsx
```diff
@@ -0,0 +1,209 @@
...
+                                    header: "Mode of Transport",
+                                    cells: [
+                                        previousConsequenceInformation.modeOfTransport,
+                                        <Link key={"mode-of-transport"} className="govuk-link" href="/add-consequence">
+                                            Change
+                                        </Link>,
+                                    ],
+                                },
+                                {
+                                    header: "Consequence Type",
```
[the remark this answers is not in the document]
alainajaffer: done

site/pages/create-consequence-operator.tsx
alainajaffer: done

site/constants/index.ts
alainajaffer: done

site/pages/create-consequence-operator.page.tsx
```diff
@@ -0,0 +1,222 @@
...
+export interface ConsequenceOperatorPageInputs {
+    "consequence-operator": string;
+    description: string;
+    "remove-from-journey-planners": string;
+    "disruption-delay": string;
+    "disruption-severity": string;
+    "disruption-direction": string;
+}
+
+export interface PageState {
```
RobCattonKPMG, line 29: Unsure if this interface needs exporting - just because we have one with the same name on the disruptions page. If we can get away with not exporting *either* of them, then lets do that.

If that causes issues, probably need to rename them both, the same as what we did for the props.

alainajaffer: done

site/pages/create-consequence-operator.page.tsx
```diff
@@ -0,0 +1,222 @@
...
+    const stateUpdater = (change: string, field: keyof ConsequenceOperatorPageInputs) => {
+        updateConsequenceOperatorPageStateForInput(field, change);
+    };
+
+    return (
+        <BaseLayout title={title} description={description}>
+            <form action="/api/createConsequenceOperator" method="post">
+                <>
+                    <div className="govuk-form-group">
+                        <h1 className="govuk-heading-xl">Add a Consequence</h1>
```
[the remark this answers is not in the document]
RobCattonKPMG: never mind - UI will be changed in future.

RobCattonKPMG (approved):

site/pages/create-consequence-operator.page.tsx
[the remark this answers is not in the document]
alainajaffer: ignored based on discussion

PR #29, opened 2023-03-13
feat: DISRUPTION-46 Validation

lj-kpmg:
- Sets up unified validation on the client and server using [zod](https://zod.dev/) schemas
- Updates form elements to support using zod schemas for validation
- Updates the /create-disruption and /type-of-consequence pages along with their APIs to use new validation approach
- Links up pages, /type-of-consequence page now links to the individual consequence pages


site/pages/api/create-disruption.api.ts
```diff
@@ -1,119 +1,44 @@
...
-        "publish-end-date": publishEndDate ? getDateTime(publishEndDate).toString() : publishEndDate,
-        "publish-start-time": publishStartTime,
-        "publish-end-time": publishEndTime,
-        "publish-no-end-date-time": publishIsNoEndDateTime,
-        "disruption-repeats": formFields["disruption-repeats"],
-    };
-
-    setCookieOnResponseObject(COOKIES_DISRUPTION_INFO, JSON.stringify(disruptionData), res);
+    try {
+        const validatedBody = createDisruptionsSchemaRefined.parse(req.body);
```
RobCattonKPMG, line 14: Is there something we gain from using .parse and catching the error as opposed to using safeParse and checking the result? Something like:

const validationResult = createDisruptionsSchemaRefined.parse(req.body);

if (!validationResult.success) {
 <failure case code>
} else {
<success case code>
}

And then we can wrap the whole thing in a try catch (similar to what's done on CFDS API's) as per the norm.

There might be something I'm not aware of which means we want to use the normal parse, just thought I'd ask.

site/pages/api/create-disruption.api.ts
```diff
@@ -1,119 +1,44 @@
...
+        setCookieOnResponseObject(COOKIES_DISRUPTION_INFO, JSON.stringify(validatedBody), res);
+        setCookieOnResponseObject(COOKIES_DISRUPTION_ERRORS, "", res, 0);
```
RobCattonKPMG, line 16: Is this clearing a cookie? I wonder if it would be useful to have a function that just takes the cookie name as an argument, and it clears it. Just because we will be reusing this code a few places I think.

site/interfaces/index.ts
```diff
@@ -1,10 +1,10 @@
 import { NextApiRequest } from "next";
+import { z } from "zod";
 import { ServerResponse } from "http";
-import { ConsequenceType, TransportMode } from "../constants/enum";
 
 export interface ErrorInfo {
     errorMessage: string;
-    id: string;
+    id: string | number;
```
alainajaffer, line 7: when would id be a number?

site/pages/type-of-consequence.test.tsx
```diff
@@ -1,19 +1,19 @@
...
 const withErrors: ErrorInfo[] = [
     { id: "consequenceType", errorMessage: "Select a consequence type" },
     { id: "modeOfTransport", errorMessage: "Select a mode of transport" },
 ];
-const withInputs: AddConsequenceProps = {
-    modeOfTransport: TransportMode.bus,
-    consequenceType: ConsequenceType.networkWide,
+const withInputs: ConsequenceTypePageInputs = {
+    modeOfTransport: VehicleMode.bus,
+    consequenceType: "networkWide",
```
alainajaffer, line 16: Why was this hardcoded and the modeOfTransport was not?

site/pages/api/type-of-consequence.test.ts
```diff
@@ -1,26 +1,27 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    setCookieSpy.mockImplementation(() => {});
     afterEach(() => {
         vi.resetAllMocks();
     });
 
-    it("should redirect back to next page when all required inputs are passed", () => {
+    it("should redirect to operator consequence page when 'Operator wide' selected", () => {
         const disruptionData = {
             modeOfTransport: "bus",
```
alainajaffer, line 26: why is this not using a type?

site/pages/create-disruption.page.tsx
```diff
@@ -1,7 +1,8 @@
 import { NextPageContext } from "next";
 import { parseCookies } from "nookies";
 import { ReactElement, useState } from "react";
-import { inspect } from "util";
+import { z } from "zod";
```
RobCattonKPMG, line 4: When you run this locally and submit the page with no inputs, no errors are shown for the end date or end time, even though the checkbox was not checked. Is this intended?

RobCattonKPMG: Also, a few of the inputs arent clearing their errors when they are filled in after this - unsure if this was intended.

site/pages/type-of-consequence.test.tsx
lj-kpmg: consequenceType is a union of strings rather than an enum so it can only be one of "networkWide", "operatorWide" etc.

site/pages/create-disruption.page.tsx
lj-kpmg: It looks like this might be because zod only runs refinements if the initial schema is valid, so it isn't getting to that check, we can either leave it as and potentially have users receive error messages twice or we could make the initial schema a partial and have a refinement check which tests to make sure all the required fields are there

lj-kpmg: I'll have a look at the errors that aren't being cleared

RobCattonKPMG (approved):

PR #30, opened 2023-03-14
feat: Create a new disruption page - validity repeat

alainajaffer:
Adding validity table and removing appropriate fields as per wireframes.

Note that this is only a front end / UI ticket and the changes made to the api will be corrected in a backend ticket (changed were only made to allow tsc checks to pass)

### DIFF
--- a/site/components/form/DateSelector.tsx
+++ b/site/components/form/DateSelector.tsx
@@ -14,6 +14,7 @@ interface DateSelectorProps<T> extends FormBase<T> {
     hiddenHint?: string;
     disablePast: boolean;
     reset?: boolean;
+    showError?: boolean;
 }
 
 const inputBox = <T extends object>(
@@ -78,6 +79,7 @@ const DateSelector = <T extends object>({
     disablePast,
     stateUpdater,
     reset = false,
+    showError = false,
 }: DateSelectorProps<T>): ReactElement => {
     const [dateValue, setDateValue] = useState<Date | null>(!!disabled || !value ? null : new Date(value));
     const [errors, setErrors] = useState<ErrorInfo[]>(initialErrors);
@@ -94,6 +96,17 @@ const DateSelector = <T extends object>({
             setDateValue(null);
         }
     }, [disabled]);
+
+    useEffect(() => {
+        if (showError) {
+            setErrors([
+                {
+                    id: inputId,
+                    errorMessage,
+                },
+            ]);
+        }
+    }, [showError, errorMessage, inputId]);
 
     return (
         <FormGroupWrapper errorIds={[inputId]} errors={errors}>
--- a/site/components/form/DateSelector.tsx
+++ b/site/components/form/DateSelector.tsx
@@ -13,6 +13,7 @@ interface DateSelectorProps<T> extends FormBase<T> {
     disabled: boolean;
     hiddenHint?: string;
     disablePast: boolean;
+    reset?: boolean;
 }
 
 const inputBox = <T extends object>(
@@ -76,9 +77,16 @@ const DateSelector = <T extends object>({
     optional = false,
     disablePast,
     stateUpdater,
+    reset = false,
 }: DateSelectorProps<T>): ReactElement => {
     const [dateValue, setDateValue] = useState<Date | null>(!!disabled || !value ? null : new Date(value));
     const [errors, setErrors] = useState<ErrorInfo[]>(initialErrors);
+
+    useEffect(() => {
+        if (reset) {
+            setDateValue(null);
+        }
+    }, [reset]);
 
     useEffect(() => {
         if (disabled) {
--- a/site/components/form/TimeSelector.tsx
+++ b/site/components/form/TimeSelector.tsx
@@ -7,6 +7,7 @@ interface TimeSelectorProps<T> extends FormBase<T> {
     disabled: boolean;
     hint?: string;
     reset?: boolean;
+    showError?: boolean;
 }
 
 const TimeSelector = <T extends object>({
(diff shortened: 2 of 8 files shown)


site/pages/create-disruption.page.tsx
```diff
@@ -65,6 +64,63 @@ const CreateDisruption = ({ inputs }: CreateDisruptionProps): ReactElement => {
         });
     };
 
+    const addValidity = (e: SyntheticEvent) => {
+        e.preventDefault();
+        setValidityClicked(true);
```
alvak1, line 69: The `validityClicked` is used to show errors when required and should be set to true when one or more of the validity inputs are empty

site/pages/create-disruption.page.tsx
```diff
@@ -65,6 +64,63 @@ const CreateDisruption = ({ inputs }: CreateDisruptionProps): ReactElement => {
...
+                    ...pageState.inputs,
+                    validity: [...pageState.inputs.validity, validityToAdd],
+                    "disruption-start-date": "",
+                    "disruption-end-date": "",
+                    "disruption-start-time": "",
+                    "disruption-end-time": "",
+                },
+                errors: pageState.errors,
+            });
+            setValidityClicked(false);
```
alvak1, line 90: This line would not be required if the `validityClicked` variable is set to true in the else block

site/pages/create-disruption.page.tsx
alainajaffer: done

site/pages/create-disruption.page.tsx
alainajaffer: It will if you have a case where you have an empty field and it triggers the errors but then you enter valid information without clicking outside (so the onBlur is not triggered). To make the errors go away and reset the state this is needed

site/components/form/DateSelector.tsx
```diff
@@ -76,17 +78,36 @@ const DateSelector = <T extends object>({
...
+    showError = false,
 }: DateSelectorProps<T>): ReactElement => {
     const [dateValue, setDateValue] = useState<Date | null>(!!disabled || !value ? null : new Date(value));
     const [errors, setErrors] = useState<ErrorInfo[]>(initialErrors);
 
+    useEffect(() => {
+        if (reset) {
+            setDateValue(null);
+        }
+    }, [reset]);
```
RobCattonKPMG, line 91: It looks like reset comes to inputBox as a boolean, which we default to false, and it never changes (because we dont use state for it). As this is the case, I don't think we need a useEffect because that will only happen on first render, and whenever reset changes (as its a dependency - but it wont ever change as its not stateful).

I think it would make more sense to have it in this line:

`const [dateValue, setDateValue] = useState<Date | null>(!!disabled || !value || !!reset ? null : new Date(value));`

Also, regarding the `new Date(value)` part, we have a getDate function we use dayJs inside of, would that be more appropriate here?

site/components/form/DateSelector.tsx
```diff
@@ -76,17 +78,36 @@ const DateSelector = <T extends object>({
...
+    useEffect(() => {
+        if (showError) {
+            setErrors([
+                {
+                    id: inputId,
+                    errorMessage,
+                },
+            ]);
+        }
+    }, [showError, errorMessage, inputId]);
```
RobCattonKPMG, line 109: Similar to the comment above, I don't think any of these dependencies are stateful and will change - so I'm not sure useEffect is appropriate unless I'm misunderstanding something.

site/components/form/TimeSelector.tsx
```diff
@@ -20,17 +22,36 @@ const TimeSelector = <T extends object>({
...
+    reset = false,
+    showError = false,
 }: TimeSelectorProps<T>): ReactElement => {
     const [errors, setErrors] = useState<ErrorInfo[]>(initialErrors);
     const [inputValue, setInputValue] = useState(value);
 
+    useEffect(() => {
+        if (reset) {
+            setInputValue("");
+        }
```
RobCattonKPMG, line 34: Same comments as above for this component.

site/pages/api/create-disruption.test.ts
```diff
@@ -265,14 +256,8 @@ describe("createDisruption", () => {
...
         const errors: ErrorInfo[] = [
-            {
-                id: "some-error-id",
-                errorMessage: "The Start date and time should be past the current date and time",
-            },
-            {
-                id: "some-error-id",
-                errorMessage: "The End date and time should be past the current date and time",
-            },
+            { id: "some-error-id", errorMessage: "The Start date and time should be past the current date and time" },
```
RobCattonKPMG, line 259: Is this the current functionality? We should be allowing users to enter start dates that are in the past, confirmed with Steve.

RobCattonKPMG: @alainajaffer I think we probably need to raise a bug rather than fix it in your PR, unless you feel its an easy change.

site/components/form/DateSelector.tsx
```diff
@@ -76,17 +78,36 @@ const DateSelector = <T extends object>({
...
+    }, [reset]);
+
     useEffect(() => {
         if (disabled) {
             setErrors([]);
             setDateValue(null);
         }
     }, [disabled]);
 
+    useEffect(() => {
```
alvak1, line 100: Selecting the date using date picker after an error does not remove the error message

site/pages/api/create-disruption.test.ts
alainajaffer: this was only changed to make sure my ticket builds as we removed some of the disruption inputs. This isn't an api ticket and will need a ticket to fix the api

PR #36, opened 2023-03-21
feat: disruptions dashboard

RobCattonKPMG:


site/data/dynamo.ts
```diff
@@ -0,0 +1,30 @@
...
+import { DynamoDBDocumentClient, QueryCommand } from "@aws-sdk/lib-dynamodb";
+import { ptSituationElementSchema } from "@create-disruptions-data/shared-ts/siriTypes.zod";
+import logger from "../utils/logger";
+
+export const getDdbClient = (): DynamoDBClient => new DynamoDBClient({ region: "eu-west-2" });
+
+export const getDdbDocumentClient = (ddbClient = getDdbClient()): DynamoDBDocumentClient =>
+    DynamoDBDocumentClient.from(ddbClient);
+
+export const getDisruptionsDataFromDynamo = async (ddbDocClient: DynamoDBDocumentClient, tableName: string) => {
```
lj-kpmg, line 11: On second thoughts it might be better if we just create an instance of the doc client outside of this function but in this file, then we wouldn't need to pass it through

lj-kpmg, line 11: We could also just define tableName in this file since it'll always be the same table

RobCattonKPMG: Will update now

site/data/dynamo.ts
```diff
@@ -0,0 +1,27 @@
...
+            TableName: process.env.TABLE_NAME as string,
+            KeyConditionExpression: "PK = :i",
+            ExpressionAttributeValues: {
+                ":i": "1",
+            },
+        }),
+    );
+
+    return dbData.Items?.map((item) => {
+        delete item.PK;
```
lj-kpmg, line 22: Looks like zod does remove any keys that aren't in the schema after parsing so we shouldn't need to bother with deleting these

site/pages/dashboard.test.tsx
```diff
@@ -0,0 +1,229 @@
+/* eslint-disable @typescript-eslint/no-unsafe-argument */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import renderer from "react-test-renderer";
+import { describe, it, expect, vi, afterEach } from "vitest";
+import Dashboard, { DashboardDisruption, getServerSideProps } from "./dashboard.page";
+import * as dynamo from "../data/dynamo";
+import { databaseData } from "../testData/mockData";
+
+const getDisruptionsSpy = vi.spyOn(dynamo, "getDisruptionsDataFromDynamo");
```
lj-kpmg, line 9: Might be worth adding `vi.mock("../data/dynamo");` in above this, just so if we had a test which didn't mock the returned value, it wouldn't try to actually connect to dynamo

site/pages/dashboard.page.tsx
```diff
@@ -0,0 +1,187 @@
...
+                summary: entry.Summary,
+                validityPeriod: {
+                    startTime: entry.ValidityPeriod.StartTime,
+                    endTime: entry.ValidityPeriod.EndTime || null,
+                },
+            };
+        });
+
+        const liveDisruptions: DashboardDisruption[] = [];
+        const upComingDisruptions: DashboardDisruption[] = [];
```
lj-kpmg, line 144: upcomingDisruptions

site/data/dynamo.ts
```diff
@@ -0,0 +1,27 @@
...
+import logger from "../utils/logger";
+
+const ddbDocClient = DynamoDBDocumentClient.from(new DynamoDBClient({ region: "eu-west-2" }));
+
+export const getDisruptionsDataFromDynamo = async () => {
+    logger.info("Getting disruptions data from DynamoDB table...");
+
+    const dbData = await ddbDocClient.send(
+        new QueryCommand({
+            TableName: process.env.TABLE_NAME as string,
```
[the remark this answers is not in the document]
RobCattonKPMG: Done

site/data/dynamo.ts
```diff
@@ -0,0 +1,27 @@
...
+                ":i": "1",
+            },
+        }),
+    );
+
+    return dbData.Items?.map((item) => {
+        delete item.PK;
+        delete item.SK;
+
+        return ptSituationElementSchema.parse(item);
```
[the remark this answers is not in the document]
RobCattonKPMG: Ok, will adjust

site/pages/dashboard.page.tsx
RobCattonKPMG: Ah, upcoming is one word!

lj-kpmg (approved):

PR #37, opened 2023-03-23
Publish disruption to dynamo

lj-kpmg:
- Uploads disruption data to dynamodb to complete steel thread journey
- Updates library used to convert json to xml
- Updated consequence schemas to work as a discriminated union in order to make the parsing easier
- Added Siri Consequence types to shared-ts folder
- Handle multiple validity periods on dashboard
- Display consequence data on review page
- Fixes issue with accordion not always rendering properly


site/data/dynamo.ts
```diff
@@ -33,3 +34,19 @@ export const getDisruptionsDataFromDynamo = async () => {
         return parsedItem.data;
     }).filter(notEmpty);
 };
+
+export const insertPublishedDisruptionIntoDynamo = async (disruption: PtSituationElement) => {
+    logger.info("Inserting disruptions data into DynamoDB table...");
```
RobCattonKPMG, line 39: Trying to think if anything would be useful to log out here, I don't know if the SituationNumber would be helpful when reading the logs as we wont know it would we?

site/pages/api/publish.api.ts
```diff
@@ -0,0 +1,179 @@
...
+        const { [COOKIES_DISRUPTION_INFO]: disruptionInfo, [COOKIES_CONSEQUENCE_INFO]: consequenceInfo } = parseCookies(
+            { req },
+        );
+
+        const parsedDisruptionInfo = createDisruptionSchema.safeParse(JSON.parse(disruptionInfo));
+        const parsedConsequenceInfo = consequenceSchema.safeParse(JSON.parse(consequenceInfo));
+
+        if (!parsedDisruptionInfo.success || !parsedConsequenceInfo.success) {
+            redirectTo(res, CREATE_DISRUPTION_PAGE_PATH);
+            return;
```
RobCattonKPMG, line 31: Do we need a TODO here over how we're going to handle this issue? I wonder if we should be throwing something here, long term.

site/pages/api/publish.api.ts
```diff
@@ -0,0 +1,179 @@
...
+                ReasonType: "PersonnelReason",
+                PersonnelReason: reason,
+            };
+        } else {
+            completePtSituationElement = {
+                ...ptSituationElement,
+                ReasonType: "EquipmentReason",
+                EquipmentReason: reason,
+            };
+        }
```
RobCattonKPMG, line 160: Not that this is a big deal, but I wonder if a switch statement is more appropriate for this?

site/pages/api/publish.api.ts
```diff
@@ -0,0 +1,179 @@
...
+
+        await insertPublishedDisruptionIntoDynamo(completePtSituationElement);
+
+        cleardownCookies(req, res);
+
+        redirectTo(res, "/dashboard");
```
RobCattonKPMG, line 166: Think we need a return after this.

site/pages/api/publish.api.ts
lj-kpmg: I don't know if there's much that would be worth doing in this situation, if a user has gone through the journey then this should never happen so it's really only to catch requests that aren't going via the expected route. I can add a warn in at least so we could see in the logs if it's happening.

RobCattonKPMG: Yeah, that makes me think we should be throwing an error so at least the user knows something didnt go as planned, and can contact helpdesk.

lj-kpmg: I've updated it to throw an error which will take them to the error page which mentions how to contact the help desk

shared-ts/siriTypes.zod.ts
```diff
@@ -50,27 +52,93 @@ export const situationElementRefSchema = z.object({
...
+            }),
+            Advice: z.object({
+                Details: z.string(),
+            }),
+            Blocking: z.object({
+                JourneyPlanner: z.boolean(),
+            }),
+            Delays: z
+                .object({
+                    Delay: z.string().regex(/^PT\d+M$/),
```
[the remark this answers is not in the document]
lj-kpmg: Done

RobCattonKPMG (approved):

PR #38, opened 2023-03-24
feat: Create Consequence Stop Page

alainajaffer:
As per wireframes the page has a dropdown, which gets data from the reference service.


site/pages/create-consequence-stops.page.tsx
```diff
@@ -0,0 +1,308 @@
...
+
+const CreateConsequenceStops = ({
+    inputs,
+    previousConsequenceInformation,
+}: CreateConsequenceProps<ConsequenceStopsPageInputs>): ReactElement => {
+    const [pageState, setPageState] = useState<PageState<Partial<ConsequenceStopsPageInputs>>>(inputs);
+    const stateUpdater = getStateUpdater(setPageState, pageState);
+    const [selected, setSelected] = useState<SingleValue<Stop>>(null);
+
+    const getOptionLabel = (e: Stop) => {
```
RobCattonKPMG, line 43: could we use 'stop' instead of e?

site/pages/create-consequence-stops.page.tsx
```diff
@@ -0,0 +1,308 @@
...
+            addStop(value);
+        }
+        setSelected(null);
+    };
+
+    const production = process.env.NODE_ENV === "production";
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const loadOptions = async (inputValue: string, _callback: (options: Stop[]) => void) => {
+        if (inputValue && inputValue.length >= 3) {
```
RobCattonKPMG, line 67: Could input value be undefined? If so it would probably make sense for the type on line 66 to be string | undefined

RobCattonKPMG: If not, we dont need to check if (input value)

site/pages/create-consequence-stops.page.tsx
```diff
@@ -0,0 +1,308 @@
...
+            const searchApiUrl = !production
+                ? "https://api.test.ref-data.dft-create-data.com/v1/stops?adminAreaCode=099"
+                : `https://api.${process.env.NODE_ENV}.ref-data.dft-create-data.com/v1/stops?adminAreaCode=099`;
+            const limit = 10;
+            const queryAdder = searchApiUrl.indexOf("?") === -1 ? "?" : "&";
```
RobCattonKPMG, line 72:
```suggestion
            const queryAdder = !searchApiUrl.includes("?") ? "?" : "&";
```

site/pages/create-consequence-stops.page.tsx
```diff
@@ -0,0 +1,308 @@
...
+            if (data) {
+                return data;
+            }
+        }
+        return [];
+    };
+
+    const removeStop = (e: SyntheticEvent, index: number) => {
+        e.preventDefault();
+        if (pageState.inputs.stopsImpacted) {
```
RobCattonKPMG, line 85: Is stopsImpacted ever undefined or is it an empty array? The reason I ask is because [] === true

site/pages/create-consequence-stops.page.tsx
alainajaffer: done

site/pages/create-consequence-stops.page.tsx
alainajaffer: done

site/pages/create-consequence-stops.page.tsx
alainajaffer: array so I added a length check

site/schemas/create-consequence-stops.schema.ts
```diff
@@ -0,0 +1,130 @@
...
+            invalid_type_error: "Nptg locality code must be a string",
+        })
+        .optional(),
+    localityName: z
+        .string({
+            required_error: "Locality name is required",
+            invalid_type_error: "Locality name must be a string",
+        })
+        .optional(),
+    parentLocalityName: z
```
RobCattonKPMG, line 49: All of these error messages need to be changed if they reference a 'string' - its not a term users will know.

site/schemas/create-consequence-stops.schema.ts
```diff
@@ -0,0 +1,130 @@
+import { Severity } from "@create-disruptions-data/shared-ts/enums";
+import { z } from "zod";
+import { setZodDefaultError, zodTimeInMinutes } from "../utils";
+
+export const stopsImpactedSchema = z.object({
+    id: z.number({
```
lj-kpmg, line 6: We should only need to define the stop attributes we actually need here, zod will remove any fields that aren't in the schema when we parse it. SIRI needs the commonName, atcoCode, latitude and longitude

site/schemas/create-consequence-stops.schema.ts
```diff
@@ -0,0 +1,130 @@
...
+        required_error: "Id is required",
```
lj-kpmg, line 7: I wouldn't bother overriding the errors if they're not being displayed to the user, zod will return an error with more detail by default

site/schemas/create-consequence-stops.schema.ts
```diff
@@ -0,0 +1,130 @@
...
+        .string({
+            required_error: "Status is required",
+            invalid_type_error: "Status must be a string",
+        })
+        .optional(),
+});
+
+export type Stop = z.infer<typeof stopsImpactedSchema>;
+
+export const createConsequenceStopsSchema = z.object({
```
lj-kpmg, line 97: In my last ticket I moved the consequence schemas into a shared consequence.schema.ts so that we can create a discriminated union of all the different consequence schemas, the contents of this file can be moved into there and we can make use of the baseConsequenceSchema rather than redefine the shared properties

site/pages/create-consequence-stops.page.tsx
```diff
@@ -0,0 +1,318 @@
...
+        setSelected(null);
+    };
+
+    const production = process.env.NODE_ENV === "production";
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const loadOptions = async (inputValue: string, _callback: (options: Stop[]) => void) => {
+        if (inputValue.length >= 3) {
+            const searchApiUrl = !production
+                ? "https://api.test.ref-data.dft-create-data.com/v1/stops?adminAreaCode=099"
```
lj-kpmg, line 69: The query param is called `adminAreaCodes` rather than `adminAreaCode`, this won't currently be filtering by area

site/pages/create-consequence-stops.page.tsx
```diff
@@ -0,0 +1,318 @@
...
+                : `https://api.${process.env.NODE_ENV}.ref-data.dft-create-data.com/v1/stops?adminAreaCode=099`;
```
lj-kpmg, line 70: `NODE_ENV` will be production for all deployed environments so this won't work, ideally we'd pass through the api url as an env var in the stack file, that way we can use the stage variable to define the url

site/pages/create-consequence-stops.page.tsx
```diff
@@ -0,0 +1,318 @@
...
+            const limit = 10;
+            const queryAdder = !searchApiUrl.includes("?") ? "?" : "&";
+            const fetchURL = `${searchApiUrl}${queryAdder}search=${inputValue}&limit=${limit}`;
```
lj-kpmg, line 73: `limit` isn't going to be doing anything here, I think we can probably simplify this to something like:
```ts
const searchApiUrl = `${API_BASE_URL}/stops?adminAreaCodes=${ADMIN_AREA_CODE}&search=${inputValue}`
```
we shouldn't need the `queryAdder`

site/pages/_document.page.tsx
```diff
@@ -16,6 +16,11 @@ const generateCsp = (): { csp: string; nonce: string } => {
     const csp = `
         default-src 'self';
         script-src 'self' 'nonce-${nonce}' 'strict-dynamic' ${production ? "" : "'unsafe-eval' 'unsafe-inline'"};
+        connect-src 'self'  ${
+            !production
+                ? "https://api.test.ref-data.dft-create-data.com/"
```
lj-kpmg, line 21: Same thing here around using an env var for the api url

site/pages/create-consequence-stops.page.tsx
```diff
@@ -0,0 +1,318 @@
...
+        ) {
+            addStop(value);
+        }
+        setSelected(null);
+    };
+
+    const production = process.env.NODE_ENV === "production";
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const loadOptions = async (inputValue: string, _callback: (options: Stop[]) => void) => {
```
lj-kpmg, line 66: Can we not just get rid of _callback here if it's not being used?

site/pages/create-consequence-stops.page.tsx
```diff
@@ -0,0 +1,318 @@
...
+            return `${stop.commonName} ${stop.atcoCode}`;
+        } else {
+            return "";
+        }
+    };
+
+    const handleChange = (value: SingleValue<Stop>) => {
+        if (
+            !pageState.inputs.stopsImpacted ||
+            pageState.inputs.stopsImpacted.filter((data) => data.id === value?.id).length === 0
```
lj-kpmg, line 56: We can use a .some rather than a .filter here:
```ts
if (!pageState.inputs.stopsImpacted ||
            pageState.inputs.stopsImpacted.some((data) => data.id === value?.id))
```

alainajaffer: done

site/pages/create-consequence-stops.page.tsx
```diff
@@ -0,0 +1,318 @@
...
+                            displaySize="l"
+                            defaultDisplay="Select severity"
+                            selectValues={DISRUPTION_SEVERITIES}
+                            stateUpdater={stateUpdater}
+                            value={pageState.inputs.disruptionSeverity}
+                            initialErrors={pageState.errors}
+                            schema={createConsequenceStopsSchema.shape.disruptionSeverity}
+                        />
+
+                        <button className="govuk-button mt-8" data-module="govuk-button">
```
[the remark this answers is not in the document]
alainajaffer: done

site/pages/_document.page.tsx
alainajaffer: done

site/pages/create-consequence-stops.page.tsx
alainajaffer: done

site/pages/create-consequence-stops.page.tsx
alainajaffer: done

site/pages/create-consequence-stops.page.tsx
alainajaffer: done

site/schemas/create-consequence-stops.schema.ts
alainajaffer: done

site/schemas/create-consequence-stops.schema.ts
alainajaffer: done

site/schemas/create-consequence-stops.schema.ts
alainajaffer: done

site/pages/create-consequence-stops.page.tsx
lj-kpmg: What error do you get? I've just tried removing it from here and from the type and it seems to build ok

site/pages/review-disruption.test.tsx
```diff
@@ -55,7 +55,7 @@ const previousConsequencesInformation: Consequence[] = [
         disruptionDelay: "12",
         disruptionDirection: "outbound",
```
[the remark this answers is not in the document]
alainajaffer: done

site/pages/create-consequence-stops.page.tsx
```diff
@@ -0,0 +1,313 @@
...
+    const getStopRows = () => {
+        if (pageState.inputs.stops) {
+            return pageState.inputs.stops.map((stop, i) => ({
+                cells: [
+                    stop.commonName && stop.indicator && stop.atcoCode
+                        ? `${stop.commonName} ${stop.indicator} ${stop.atcoCode}`
+                        : `${stop.commonName} ${stop.atcoCode}`,
+                    <button
+                        id={`remove-stop-${i + 1}`}
+                        key={`remove-stop-${i + 1}`}
```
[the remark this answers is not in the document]
alainajaffer: done

site/pages/api/create-consequence-stops.test.ts
```diff
@@ -0,0 +1,129 @@
...
+    disruptionSeverity: Severity.severe,
+    vehicleMode: VehicleMode.bus,
+    consequenceType: "stops",
+};
+
+describe("create-consequence-stops API", () => {
+    const writeHeadMock = vi.fn();
+    vi.mock("../../utils/apiUtils", async () => ({
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+        ...((await vi.importActual("../../utils/apiUtils")) as object),
```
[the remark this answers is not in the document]
alainajaffer: done

lj-kpmg (approved):

PR #49, opened 2023-04-04
CSRF check on form submission

lj-kpmg:


site/pages/cookies.page.tsx
```diff
@@ -139,7 +137,7 @@ export const getServerSideProps = (ctx: NextPageContext): { props: CookiePrefere
 
     const trackingDefaultValue = tracking ? "on" : "off";
 
-    return { props: { settingsSaved, trackingDefaultValue, csrfToken } };
```
RobCattonKPMG, line 142: How come this page doesnt need it?

site/components/form/CsrfForm.tsx
```diff
@@ -9,8 +9,8 @@ interface CsrfFormProps {
 }
```
RobCattonKPMG, line 9: Could we make csrfToken string or undefined, and default it in the component to an empty string? So that we dont need to always do csrfToken={csrfToken || ""}

site/pages/cookies.page.tsx
lj-kpmg: The token is injected in _app.page.tsx so it doesn't need to be passed through in the getServerSideProps

RobCattonKPMG (approved):

PR #52, opened 2023-04-04
Date picker error reset fix

alvak1:
Fix for errors to be reset when the date is selected with pre-populated errors

### DIFF
--- a/site/components/form/DateSelector.tsx
+++ b/site/components/form/DateSelector.tsx
@@ -41,7 +41,11 @@ const inputBox = <T extends object>(
                 {...inputProps}
                 disabled={disabled}
                 placeholder={disabled ? "N/A" : "DD/MM/YYYY"}
-                onBlur={(e) => handleBlur(e.target.value, inputName, stateUpdater, setErrors, schema, disabled)}
+                onBlur={(e) => {
+                    if (!e.target.value) {
+                        handleBlur(e.target.value, inputName, stateUpdater, setErrors, schema, disabled);
+                    }
+                }}
             />
         </div>
         {InputProps?.endAdornment}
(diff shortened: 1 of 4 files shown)


site/components/form/DateSelector.tsx
```diff
@@ -25,27 +25,24 @@ const inputBox = <T extends object>(
...
-                    id={`${inputId}-input`}
-                    type="text"
-                    ref={inputRef}
-                    {...inputProps}
-                    disabled={disabled}
-                    placeholder={disabled ? "N/A" : "DD/MM/YYYY"}
-                    onBlur={(e) => handleBlur(e.target.value, inputName, stateUpdater, setErrors, schema, disabled)}
-                />
-            </FormElementWrapper>
+            <input
```
RobCattonKPMG, line 35: Found a couple of bugs when testing this locally:

If you enter a partially correct date, but dont enter it fully correctly, when you navigate away from the box it clears the box. This is new behaviour and isn't what we want.

If the box is empty and has an error, if you click into the box and type, the first character you type is ignored and the error disappears

RobCattonKPMG: @alvak1 looks to be working, but it has regressed so that if you enter 01/01/202 as an input (or delete the last character of a correct date input) then there is no error message when you click away.

alvak1: When the errors are set or reset the child elements of FormElementWrapper returns different objects. When the returned object is different to what was pervious displayed, the value of the text box is reset as its rendered as part of a new object.

The current fix deletes the entire date when we try to delete the last character. The error message which previously would be remove the error message when a character is entered, will remove the error only when valid date is entered.

alvak1: Reverted the code to one of earlier fixes where the errors are reset

RobCattonKPMG (approved):

PR #56, opened 2023-04-06
feat: Polygon for stops page

alainajaffer:
Note my reference service one needs to go in first for this to work but this can be tested running reference service locally and replacing the API_BASE_URL

https://github.com/Department-for-Transport-Disruptions/reference-data-service/pull/8


site/pages/create-consequence-stops.page.tsx
```diff
@@ -322,10 +334,12 @@ export const getServerSideProps = (ctx: NextPageContext): { props: CreateConsequ
         return;
     }
 
+    const allStops = await fetchStops();
```
lj-kpmg, line 337: Is this still needed now that we're calling the API on demand with the polygon?

lj-kpmg (approved):

PR #62, opened 2023-04-13
disruption validity repetition which allows daily and weekly disruptions to be added

alvak1:
Access page `/create-disruption` and proceed with adding disruptions


site/pages/create-consequence-services/__snapshots__/create-consequence-services.test.tsx.snap
```diff
@@ -626,12 +626,14 @@ exports[`pages > CreateConsequenceServices > should render correctly with errors
                 <div
                   aria-describedby="removeFromJourneyPlanners-error"
                   className="govuk-radios govuk-radios--error"
+                  data-module="govuk-radios"
```
alainajaffer, line 629: When I choose daily and then enter all the details besides the start time then press submit the radios become disabled and default to doesn't repeat rather than what I entered for weekly - is that intended?

alainajaffer: I think it is the ordering I select weekly or daily first then enter a value then select no end date and time so then it doesn't revert back to doesn't repeat rather it just disabled the radios on the value I entered

site/pages/create-disruption/[disruptionId].page.tsx
```diff
@@ -285,6 +344,67 @@ const CreateDisruption = (props: DisruptionPageProps): ReactElement => {
                             reset={addValidityClicked}
                             schema={validitySchema.shape.disruptionNoEndDateTime}
                         />
+
+                        <Radios<DisruptionInfo>
+                            display="Does this diruption repeat"
```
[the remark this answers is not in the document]
Comment (alvak1) [reply site/pages/create-disruption/[disruptionId].page.tsx]: done

site/pages/create-consequence-services/__snapshots__/create-consequence-services.test.tsx.snap
alvak1: done

site/components/form/Radios.tsx
```diff
@@ -1,14 +1,20 @@
 import kebabCase from "lodash/kebabCase";
-import { ReactElement, useState } from "react";
+import { Fragment, ReactElement, RefObject, useEffect, useState } from "react";
```
alainajaffer, line 2: On the review page could it state Repeats [daily/weekly] until [Ending on] aswell as just the start and end date ?

site/sst-env.d.ts
```diff
@@ -1 +1,2 @@
+// eslint-disable-next-line @typescript-eslint/triple-slash-reference
```
alainajaffer, line 1: is this needed?

site/sst-env.d.ts
```diff
@@ -1 +1,2 @@
...
 /// <reference path="../.sst/types/index.ts" />
```
alainajaffer, line 2:
```suggestion
// <reference path="../.sst/types/index.ts" />
```

site/sst-env.d.ts
alvak1: Removed this file as @lj-kpmg  will be adding this file

site/sst-env.d.ts
alvak1: As per above comment this file is removed

alainajaffer: why does this need to be commented in the first place?

site/pages/dashboard.page.tsx
```diff
@@ -1,3 +1,4 @@
+import dayjs from "dayjs";
```
RobCattonKPMG, line 1: We dont want dayjs being imported to pages. We have a file called dates.ts inside the utils where all the date functions live.

site/pages/dashboard.page.tsx
```diff
@@ -217,6 +236,9 @@ export const getServerSideProps = async (): Promise<{ props: DashboardProps }> =
                     disruptionEndDate: disruption.disruptionEndDate,
```
RobCattonKPMG, line 236: This is where we should be taking the disruption.validity array, and if we have a disruption repeat, we override the last date with the date a year from now.

site/pages/dashboard.page.tsx
alvak1: done, using functions from dates.js where possible

site/pages/dashboard.page.tsx
alvak1: Updated the code to run the end date check for repeating disruptions at server side

site/pages/dashboard.page.tsx
```diff
@@ -29,8 +30,27 @@ export interface DashboardProps {
     newDisruptionId: string;
 }
 
-const mapDisruptions = (disruptions: Disruption[]) =>
-    sortDisruptionsByStartDate(disruptions).map((disruption) => {
+const mapDisruptions = (disruptions: Disruption[]) => {
+    return sortDisruptionsByStartDate(disruptions).map((disruption) => {
+        let maxEndDate: Dayjs | null = getDate().subtract(100, "years");
```
RobCattonKPMG, line 35: We dont need to add the Dayjs | null type here - it will be inferred by the IDE. 

Also, why is the maxEndDate 100 years ago?

site/pages/api/create-disruption.api.ts
```diff
@@ -26,6 +26,9 @@ export const formatCreateDisruptionBody = (body: object) => {
                 disruptionEndDate: values[2],
                 disruptionEndTime: values[3],
                 disruptionNoEndDateTime: values[4],
+                disruptionRepeats: values[5],
+                disruptionDailyRepeatsEndDate: values[6],
```
RobCattonKPMG, line 30: We just need disruptionRepeats and disruptionRepeatEndDate here

site/pages/create-disruption/[disruptionId].page.tsx
```diff
@@ -105,22 +124,43 @@ const CreateDisruption = (props: DisruptionPageProps): ReactElement => {
...
+                        : validity.disruptionWeeklyRepeatsEndDate;
+
+                return {
+                    header: `Validity period ${i + 1}`,
+                    cells: [
+                        validity.disruptionEndDate &&
+                        validity.disruptionEndTime &&
+                        !validity.disruptionNoEndDateTime ? (
+                            validity.disruptionRepeats !== "doesntRepeat" &&
+                            validity.disruptionRepeats &&
```
Comment (RobCattonKPMG) [inline site/pages/create-disruption/[disruptionId].page.tsx:140]: We should be checking if disruptionRepeats before checking if it equals something.

site/pages/dashboard.page.tsx
```diff
@@ -29,8 +30,27 @@ export interface DashboardProps {
...
+
+        disruption.validity?.forEach((validity) => {
+            const repeatsEndDate =
+                validity.disruptionRepeats === "daily" && validity.disruptionDailyRepeatsEndDate
+                    ? getFormattedDate(validity.disruptionDailyRepeatsEndDate)
+                    : validity.disruptionRepeats === "weekly" && validity.disruptionWeeklyRepeatsEndDate
+                    ? getFormattedDate(validity.disruptionWeeklyRepeatsEndDate)
+                    : validity.disruptionEndDate && validity.disruptionEndTime
+                    ? getDatetimeFromDateAndTime(validity.disruptionEndDate, validity.disruptionEndTime)
+                    : null;
```
RobCattonKPMG, line 45: I think this code needs to be doing the following: 

- map over the validity periods to see if any of them contain 'disruptionRepeats'.
- If any of them repeat, check to see if they have an end date.
- If they dont, give them a max end date (the function to get a max end date should live in the Dates.ts file - we shouldnt be doing the subtraction here).

Then, the page already gets the final end date of the final validity period and uses it as the end date to show on the screen.

site/pages/dashboard.page.tsx
alvak1: Removing Dayjs | null throws an error in line 50 `Type 'null' is not assignable to type 'Dayjs'`

For maxEndDate the initialisation value is a value in past so that when the iteration begins it is replaced by the value in the first validity

RobCattonKPMG: I dont understand what you mean here - arent we just finding if there is a repeating validity without an end date, and if there is then we're putting in a value of a year from today?

alvak1: yes and if there is no end date then we are setting the value to null cause this null value is used in `formatDisruptionsIntoRows` function to display end date or onwards text

site/pages/api/create-disruption.api.ts
alvak1: done

site/pages/dashboard.page.tsx
alvak1: The `sortDisruptionsByStartDate` method would sort the validity records as per the start date. This is how the earliest start date is obtained for the disruption.

The end date is obtained from the last validity record. This record was sorted based on start date and has not accounted for ending on date. 

If there are records with ending on date later than the end date of the last record this data is not captured because its not the last validity record. As the start date of this record with ending on date could be before the start date of the record in the last validity index

site/pages/create-disruption/[disruptionId].page.tsx
Comment (alvak1) [reply site/pages/create-disruption/[disruptionId].page.tsx:137]: done

site/pages/api/create-disruption.api.ts
```diff
@@ -26,16 +26,31 @@ export const formatCreateDisruptionBody = (body: object) => {
...
                 disruptionNoEndDateTime: values[4],
+                disruptionRepeats: values[5],
+                disruptionRepeatsEndDate: values[6],
             };
         });
 
+    const disruptionRepeatsEndDate = Object.entries(body)
+        .filter((item) => item.toString().startsWith("disruptionRepeatsEndDate"))
+        .map((arr: string[]) => {
+            const [, values] = arr;
```
RobCattonKPMG, line 37: whats the reason for the ',' here?

site/pages/dashboard.page.tsx
RobCattonKPMG: I still dont understand why we need a date 100 years in the past?

RobCattonKPMG: If we find the validityPeriod that we need to set a date in, cant we just insert the endDate as a year from the start date, and job done?

site/pages/review-disruption/[disruptionId].page.tsx
```diff
@@ -65,18 +65,41 @@ const ReviewDisruption = ({ disruption, previousSocialMediaPosts, csrfToken }: R
...
+                        Repeats {validity.disruptionRepeats} until {validity.disruptionRepeatsEndDate}
+                    </>
+                ) : validity.disruptionRepeats === "weekly" ? (
+                    <>
+                        <br />
+                        Repeats every week until {validity.disruptionRepeatsEndDate}
+                    </>
+                ) : (
+                    <></>
+                );
```
Comment (RobCattonKPMG) [inline site/pages/review-disruption/[disruptionId].page.tsx:87]: What are the <br> 's for?

Comment (alvak1) [reply site/pages/review-disruption/[disruptionId].page.tsx:87]: <br/> is for line break. The text following br should be displayed in next line

site/pages/dashboard.page.tsx
RobCattonKPMG: But validity periods cant overlap, which means the end date of the last one in the array must be the latest end date?

site/pages/review-disruption/[disruptionId].page.tsx
Comment (RobCattonKPMG) [reply site/pages/review-disruption/[disruptionId].page.tsx:87]: If we just had 
```suggestion
            Repeats {validity.disruptionRepeats} until {validity.disruptionRepeatsEndDate}
```

then it would work for both daily and weekly, and we wouldnt need 2 versions.

Comment (RobCattonKPMG) [reply site/pages/review-disruption/[disruptionId].page.tsx:87]: And ideally we dont need to use <br /> - the append value just needs to be a string, and we can do the html stuff in the next part. Then we arent adding html to html.

Comment (alvak1) [reply site/pages/review-disruption/[disruptionId].page.tsx:87]: Was told that the text should be of below format. Above suggestion will not display "every week"

"Repeats {"daily"/"every week"} until {Ending on}"

RobCattonKPMG (approved):

PR #64, opened 2023-04-17
feat: Service routes on map

alainajaffer:
Note that for the select all functionality services does not have a functioning polygon - to be done in another ticket 
Doesn't have the service pop-up - blocked on that

### DIFF
--- a/site/components/map/Map.tsx
+++ b/site/components/map/Map.tsx
@@ -1,4 +1,6 @@
+import { Feature, GeoJsonProperties, Geometry } from "geojson";
 import uniqueId from "lodash/uniqueId";
+import { LineLayout, LinePaint } from "mapbox-gl";
 import {
     CSSProperties,
     Dispatch,
@@ -10,25 +12,46 @@ import {
     useEffect,
     useState,
 } from "react";
-import MapBox, { Marker, Popup, ViewState } from "react-map-gl";
+import MapBox, { Layer, Marker, Popup, Source, ViewState } from "react-map-gl";
 import { z } from "zod";
 import DrawControl, { PolygonFeature } from "./DrawControl";
 import { ADMIN_AREA_CODE, API_BASE_URL } from "../../constants";
 import { PageState } from "../../interfaces";
-import { Stop, StopsConsequence, stopSchema, stopsConsequenceSchema } from "../../schemas/consequence.schema";
+import {
+    Routes,
+    ServicesConsequence,
+    Stop,
+    StopsConsequence,
+    servicesConsequenceSchema,
+    stopSchema,
+    stopsConsequenceSchema,
+} from "../../schemas/consequence.schema";
 import { flattenZodErrors } from "../../utils";
 import { sortStops } from "../../utils/formUtils";
 interface MapProps {
     initialViewState: Partial<ViewState>;
     style: CSSProperties;
     mapStyle: string;
-    selected?: Stop[];
-    searched?: Stop[];
+    selected: Stop[];
+    searched: Stop[];
     inputId?: keyof Stop;
     showSelectAllButton?: boolean;
-    stateUpdater?: Dispatch<SetStateAction<PageState<Partial<StopsConsequence>>>>;
-    state?: PageState<Partial<StopsConsequence>>;
+    stateUpdater:
+        | Dispatch<SetStateAction<PageState<Partial<StopsConsequence>>>>
+        | Dispatch<SetStateAction<PageState<Partial<ServicesConsequence>>>>;
+    state: PageState<Partial<StopsConsequence>> | PageState<Partial<ServicesConsequence>>;
+    searchedRoutes?: Routes;
 }
+
+const lineLayout: LineLayout = {
+    "line-join": "round",
+    "line-cap": "round",
+};
+
+const lineStyle: LinePaint = {
+    "line-color": "rgba(3, 170, 238, 0.5)",
+    "line-width": 5,
+};
 
 const Map = ({
     initialViewState,
@@ -39,6 +62,7 @@ const Map = ({
     showSelectAllButton = false,
     stateUpdater = () => "",
     state,
+    searchedRoutes,
 }: MapProps): ReactElement | null => {
     const mapboxAccessToken = process.env.MAP_BOX_ACCESS_TOKEN;
     const [features, setFeatures] = useState<{ [key: string]: PolygonFeature }>({});
@@ -46,9 +70,18 @@ const Map = ({
     const [selectAll, setSelectAll] = useState<boolean>(true);
     const [popupInfo, setPopupInfo] = useState<Partial<Stop>>({});
 
+    const createLineString = (coordinates: Stop[]): Feature<Geometry, GeoJsonProperties> => ({
+        type: "Feature",
+        properties: {},
+        geometry: {
+            type: "LineString",
+            coordinates: coordinates.map((stop) => [stop.longitude, stop.latitude]),
+        },
+    });
+
     const handleMouseEnter = useCallback(
         (id: string) => {
-            const stopsOnMap = searched && selected ? [...selected, ...searched] : [];
+            const stopsOnMap = [...selected, ...searched];
             const stopInfo = stopsOnMap.find((stop) => stop.atcoCode === id);
             if (stopInfo) setPopupInfo(stopInfo);
         },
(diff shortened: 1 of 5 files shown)


site/components/map/Map.tsx
```diff
@@ -261,14 +361,137 @@ const Map = ({
         setSelectAll(!selectAll);
     };
 
+    useEffect(() => {
+        setSelectAll(true);
+    }, [searchedRoutes]);
```
RobCattonKPMG, line 366: Why are we setting selectAll to be true when searchedRoutes changes?

site/components/map/Map.tsx
```diff
@@ -261,14 +361,137 @@ const Map = ({
...
+        }
+    }, []);
+
+    const selectedService = (hoverInfo && hoverInfo.serviceId) || "";
+    const filter = useMemo(() => ["==", "serviceId", selectedService], [selectedService]);
+
+    const getSourcesInbound = useCallback(
+        (searchedRoutes: Partial<(Routes & { serviceId: number })[]>) =>
+            searchedRoutes.map((searchedRoute) =>
+                searchedRoute?.inbound ? (
```
RobCattonKPMG, line 385: Maybe change this to searchedRoute?.inbound && searchedRoute?.serviceId so that you can just do searchedRoute.serviceId below

site/components/map/Map.tsx
```diff
@@ -261,14 +361,137 @@ const Map = ({
...
+                ) : null,
+            ),
+
+        [filter],
+    );
+
+    const getSourcesOutbound = useCallback(
+        (searchedRoutes: Partial<(Routes & { serviceId: number })[]>) =>
+            searchedRoutes.map((searchedRoute) =>
+                searchedRoute?.outbound ? (
```
RobCattonKPMG, line 417: Same comment as above.

site/components/map/Map.tsx
```diff
@@ -261,14 +361,137 @@ const Map = ({
...
+                ) : null,
+            ),
+        [filter],
+    );
+
+    const getInteractiveLayerIds = useCallback(
+        () =>
+            searchedRoutes && searchedRoutes.length > 0
+                ? searchedRoutes.flatMap((sr) => {
+                      if (sr?.inbound && sr.outbound) {
```
RobCattonKPMG, line 449: Do we need sr?.inbound instead of sr.inbound? Outbound seems ok without it. We can also check for serviceId so we dont have to do ?. below

site/components/map/Map.tsx
```diff
@@ -261,14 +361,137 @@ const Map = ({
...
+                          return [
+                              `services-inbound-${sr?.serviceId || ""}`,
+                              `services-outbound-${sr?.serviceId || ""}`,
+                          ];
+                      } else if (sr?.inbound) {
```
RobCattonKPMG, line 454: I dont think we need the else's because we have returns in each if.

site/components/map/Map.tsx
```diff
@@ -261,14 +361,137 @@ const Map = ({
...
     return mapboxAccessToken ? (
         <>
+            {selected.length === 100 ? (
+                <div className="govuk-warning-text">
+                    <span className="govuk-warning-text__icon" aria-hidden="true">
+                        !
+                    </span>
+                    <strong className="govuk-warning-text__text">
+                        <span className="govuk-warning-text__assistive">Warning</span>
+                        {`Stop selection capped at 100, ${selected.length || "0"} stops currently selected`}
```
RobCattonKPMG, line 481: I think ${selected.length || "0"} can just be selected.length as its an array thats passed in, it will always have a length.

site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx
```diff
@@ -127,7 +153,7 @@ const CreateConsequenceServices = (props: CreateConsequenceServicesProps): React
                 ],
             });
         } else {
-            if (stopToAdd) {
+            if (stopToAdd && (pageState.inputs.stops ? pageState.inputs.stops.length < 100 : true)) {
```
Comment (RobCattonKPMG) [inline site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx:156]: Should this be <= 100 as 100 is the limit?

site/pages/create-consequence-stops/[disruptionId]/[consequenceIndex].page.tsx
```diff
@@ -115,7 +115,7 @@ const CreateConsequenceStops = (props: CreateConsequenceStopsProps): ReactElemen
                 ],
             });
         } else {
-            if (stopToAdd) {
+            if (stopToAdd && (pageState.inputs.stops ? pageState.inputs.stops.length < 100 : true)) {
```
Comment (RobCattonKPMG) [inline site/pages/create-consequence-stops/[disruptionId]/[consequenceIndex].page.tsx:118]: Same question about <=100 here

site/components/map/Map.tsx
alainajaffer: Let's say I have clicked the select all button before when selecting a service - if a service changes its searchedRoutes will change. To prevent the button from staying on 'unselect all' I change reset it back to 'select all'.

site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx
Comment (alainajaffer) [reply site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx:156]: don't want to add anything if we are already at 100 - if it is 99 we want to add a stop

site/pages/create-consequence-stops/[disruptionId]/[consequenceIndex].page.tsx
Comment (alainajaffer) [reply site/pages/create-consequence-stops/[disruptionId]/[consequenceIndex].page.tsx:118]: see above

site/components/map/Map.tsx
alainajaffer: it complains about sr not being defined if i remove the sr?.

RobCattonKPMG: Do we know why Inbound might be undefined on a searchRoute but oubound is never undefined?

RobCattonKPMG: Or rather, do we know why it works for outbound and not for inbound

site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx
Comment (RobCattonKPMG) [reply site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx:156]: ah okay - just realised we have a boolean mixed with an inline if. It could be simplified to `stopToAdd && pageState.inputs.stops && pageState.inputs.stops.length < 100`

site/components/map/Map.tsx
RobCattonKPMG: Ah - I didnt realise selectAll (boolean) was being used for the text of the button. Can we just check that all the searchedRoutes are selected to decide what text to have the button as?

RobCattonKPMG: Yeah just reread the file - when the component is first rendered, selectAll is true - id expect selectAll to be true if the user has clicked it or has manually selected all the options.

site/components/map/Map.tsx
alainajaffer: done

site/components/map/Map.tsx
alainajaffer: done

site/components/map/Map.tsx
alainajaffer: done serviceId check

site/components/map/Map.tsx
alainajaffer: done

site/components/map/Map.tsx
alainajaffer: done

site/components/map/Map.tsx
RobCattonKPMG: So when the user clicks the button, they have chosen to selectAll, so selectAll would be true.

site/components/map/Map.tsx
alainajaffer: sr.outbound only becomes defined when i check inbound first unless I do sr?.outbound it seems like if one exists the other also does

site/components/map/Map.tsx
alainajaffer: it would be hard to check if all the routes are selected due to the 100 stops limitation

site/components/map/Map.tsx
RobCattonKPMG: Fair enough!

site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx
Comment (alainajaffer) [reply site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx]: done

site/components/map/Map.tsx
RobCattonKPMG: Yeah fair enough - but I do think the value should be reversed. If the user comes into the page, and they havent selected all - then selectAll should be false. Having it be true, and checking for !selectAll is a little confusing.

alainajaffer: done

RobCattonKPMG (approved):

PR #66, opened 2023-04-21
Update to overlap logic 

alvak1:
Validity Start date/time and end date/time should be considered to check for overlapping validities. Ending on date for repeating disruptions are not part of overlap check


site/pages/api/create-disruption.test.ts
```diff
@@ -622,126 +611,4 @@ describe("create-disruption API", () => {
         );
         expect(writeHeadMock).toBeCalledWith(302, { Location: `/create-disruption/${defaultDisruptionId}` });
     });
-
-    it("should confirm that the dates are expanded as expected", () => {
```
RobCattonKPMG, line 626: Why are we deleting all these tests?

site/schemas/create-disruption.schema.ts
```diff
@@ -499,33 +499,6 @@ export const createDisruptionsSchemaRefined = createDisruptionSchema
             ];
 
             let valid = true;
```
RobCattonKPMG, line 501: What is valid used for now?

site/pages/api/create-disruption.test.ts
alvak1: The test case was for expandedValidity function. As this function is deleted the test case would redundant

site/schemas/create-disruption.schema.ts
alvak1: isValid is removed

RobCattonKPMG (approved):

PR #67, opened 2023-04-21
feature: Operator search component

RobCattonKPMG:
Added a new component to the site for searching for operators, which is used on the consequence operator page. Note that the same behaviour is expected via the API - only the first operator in the selected list will be used.


site/schemas/consequence.schema.ts
```diff
@@ -110,3 +107,20 @@ export const consequenceSchema = z.discriminatedUnion("consequenceType", [
 ]);
 
 export type Consequence = z.infer<typeof consequenceSchema>;
+
+export const operatorSchema = z.object({
+    id: z.number(),
+    nocCode: z.string(),
+    operatorPublicName: z.string(),
+    vosaPsvLicenseName: z.string().optional(),
```
alainajaffer, line 115: will we need all of these attributes?

site/testData/mockData.ts
```diff
@@ -420,3 +420,76 @@ export const ptSituationElementWithMultipleConsequences = {
...
+    },
+    {
+        id: 6,
+        nocCode: "8H",
+        operatorPublicName: "Highland Airways",
+        vosaPsvLicenseName: "Highland Airways",
+        opId: "136620",
+        pubNmId: "94334",
+        nocCdQual: "",
+        changeDate: "",
```
alainajaffer, line 489: Are all of these needed?

site/pages/create-consequence-operator/[disruptionId]/[consequenceIndex].page.tsx
```diff
@@ -76,16 +93,47 @@ const CreateConsequenceOperator = (props: CreateConsequenceOperatorProps): React
...
+                                        ],
+                                    };
+                                })}
+                            />
+                        ) : null}
+
+                        <input
+                            type="hidden"
+                            name="consequenceOperator"
+                            value={selectedOperators.length > 0 ? selectedOperators[0].nocCode : undefined}
```
Comment (alainajaffer) [inline site/pages/create-consequence-operator/[disruptionId]/[consequenceIndex].page.tsx:136]: should the else here be null?

site/schemas/consequence.schema.ts
RobCattonKPMG: Yeah I'm not sure - the API call will be changing early next week, so I figured I'd add it all and we can cut it back as needed.

site/testData/mockData.ts
RobCattonKPMG: Its just copy and pasted data that came back from the API.

site/pages/create-consequence-operator/[disruptionId]/[consequenceIndex].page.tsx
Comment (RobCattonKPMG) [reply site/pages/create-consequence-operator/[disruptionId]/[consequenceIndex].page.tsx:136]: Yeah I can change it to null.

Comment (RobCattonKPMG) [reply site/pages/create-consequence-operator/[disruptionId]/[consequenceIndex].page.tsx:136]: Ah - null is not an allowed value.

site/schemas/consequence.schema.ts
alainajaffer: ok that is fine for now

RobCattonKPMG: Its not about whats being displayed - we only need the NOC and the Operator Name (operatorPublicName) - its about what we need in the back end for when we're using the operator information.

alainajaffer (approved):

PR #68, opened 2023-04-21
bugfix: Fixing stops dropdown

alainajaffer:
The stops dropdown was not allowing a user to search and click for a stop


alvak1 (approved):
Approved

PR #71, opened 2023-04-24
filter operators by admin area

lj-kpmg:


site/sst-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference path="../.sst/types/index.ts" />
```
alainajaffer, line 1: is this needed?

alainajaffer: resolved in conversation

alainajaffer (approved):

PR #73, opened 2023-04-24
feat: Delete disruption

alainajaffer:
The button is on the review page at the moment but will be moved once @alvak1's ticket is done

All functionality and tests have been implemented

Navigate to the dashboard click on a disruption id and then delete the disruption or return (test both cases)


site/components/DeleteConfirmationPopup.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { describe, it, expect, afterEach } from "vitest";
+import DeleteConfirmationPopup from "./DeleteConfirmationPopup";
+
+describe("DeleteConfirmationPopup", () => {
+    it("should render correctly", () => {
+        const { container } = render(
+            <DeleteConfirmationPopup
+                entityName="test"
+                deleteUrl="https://test.com/api/delete"
+                cancelActionHandler={() => ""}
```
RobCattonKPMG, line 11: Should we be using vi.fn() here? I think most tests do that instead of these empty functions.

RobCattonKPMG: Also in the rest of the file

site/pages/api/delete-disruption.api.ts
```diff
@@ -0,0 +1,42 @@
...
+import { ERROR_PATH } from "../../constants";
+import { deletePublishedDisruption, getDisruptionById } from "../../data/dynamo";
+import { redirectTo, redirectToError } from "../../utils/apiUtils";
+import logger from "../../utils/logger";
+
+const deleteDisruption = async (req: NextApiRequest, res: NextApiResponse): Promise<void> => {
+    try {
+        const { query } = req;
+
+        const id = query?.id as string;
```
RobCattonKPMG, line 12:
```suggestion
        const id = query?.id as string | undefined;
```

site/pages/api/delete-disruption.api.ts
```diff
@@ -0,0 +1,42 @@
...
+
+        if (!id) {
```
RobCattonKPMG, line 14: if the user passes through the id as an array (I think thats a potential type it can be?) then we need a check inside this if I think (super niche).

site/pages/api/delete-disruption.api.ts
```diff
@@ -0,0 +1,42 @@
...
+            redirectTo(res, ERROR_PATH);
+            return;
+        }
+
+        await deletePublishedDisruption(getPtSituationElementFromDraft(disruption), id);
+
+        redirectTo(res, "/dashboard");
+        return;
+    } catch (e) {
+        if (e instanceof Error) {
```
RobCattonKPMG, line 31: Could potentially be useful to change the code so we have access to the id, and then we can log it out.

alainajaffer: added it to the error above felt it was not appropriate here

site/pages/api/delete-disruption.api.ts
alainajaffer: done if it is an array an error is thrown

site/pages/api/delete-disruption.api.ts
alainajaffer: removed no longer needed

site/components/DeleteConfirmationPopup.test.tsx
alainajaffer: done

RobCattonKPMG (dismissed):

site/pages/api/delete-disruption.api.ts
```diff
@@ -0,0 +1,44 @@
...
+
+        const disruption = await getDisruptionById(id);
+
+        if (!disruption) {
+            logger.error(`Disruption ${id} not found to delete`);
+            redirectTo(res, ERROR_PATH);
+            return;
+        }
+
+        await deletePublishedDisruption(getPtSituationElementFromDraft(disruption), id);
```
lj-kpmg, line 28: We could probably simplify this by just passing through the disruption without converting it into a PtSituationElement since all we really need to delete it is the ID and the number of consequences

site/pages/review-disruption/[disruptionId].page.tsx
```diff
@@ -43,6 +44,19 @@ const getConsequenceUrl = (type: Consequence["consequenceType"]) => {
...
+    const deleteActionHandler = (id: string, name: string): void => {
+        setPopUpState({ disruptionId: id, disruptionName: name });
+    };
+
+    const cancelActionHandler = (): void => {
+        setPopUpState(undefined);
+    };
+
+    const buildDeleteUrl = (idToDelete: string, csrfToken: string): string => {
+        return `/api/delete-disruption?id=${idToDelete}&_csrf=${csrfToken}`;
```
Comment (lj-kpmg) [inline site/pages/review-disruption/[disruptionId].page.tsx:58]: Is there a reason for passing the csrf token through here rather than using the CsrfForm? Also it might be better to pass the disruptionId through as a hidden input in the form to keep it consistent with what we're doing elsewhere on the site

site/pages/review-disruption/[disruptionId].page.tsx
```diff
@@ -413,6 +427,24 @@ const ReviewDisruption = ({ disruption, previousSocialMediaPosts, csrfToken }: R
...
+                            className="govuk-button govuk-button--warning ml-5 mt-8"
+                            data-module="govuk-button"
+                            onClick={(e) => {
+                                e.preventDefault();
+                                deleteActionHandler(disruption.disruptionId, disruption.summary);
+                            }}
+                        >
+                            Delete disruption
+                        </button>
+                        {popUpState && csrfToken ? (
```
Comment (lj-kpmg) [inline site/pages/review-disruption/[disruptionId].page.tsx:440]: I'd maybe move this outside the CsrfForm element since we should avoid nesting forms where possible

site/pages/review-disruption/[disruptionId].page.tsx
Comment (alainajaffer) [reply site/pages/review-disruption/[disruptionId].page.tsx:58]: i was doing it how fares does it but i can change the approach to this

site/pages/review-disruption/[disruptionId].page.tsx
Comment (alainajaffer) [reply site/pages/review-disruption/[disruptionId].page.tsx:440]: done

site/pages/review-disruption/[disruptionId].page.tsx
Comment (alainajaffer) [reply site/pages/review-disruption/[disruptionId].page.tsx:58]: done

site/pages/api/delete-disruption.api.ts
alainajaffer: done

lj-kpmg (approved):

PR #77, opened 2023-04-26
account for consequence indexes when deleting

lj-kpmg:


site/pages/review-disruption/[disruptionId].page.tsx
```diff
@@ -129,6 +129,10 @@ const ReviewDisruption = ({
         });
     };
 
+    const largestIndex = disruption.consequences?.reduce((p, c) =>
```
Comment (alainajaffer) [inline site/pages/review-disruption/[disruptionId].page.tsx:132]: Reduce of empty array with no initial value

alainajaffer (approved):

PR #83, opened 2023-04-28
feat:  Services polygon

alainajaffer:
Includes DISRUPTION-113 tooltip: unselected stops within polygon

### DIFF
--- a/site/components/map/ServicesMap.tsx
+++ b/site/components/map/ServicesMap.tsx
@@ -17,15 +17,14 @@ import MapBox, { Layer, Marker, Popup, Source, ViewState } from "react-map-gl";
 import { z } from "zod";
 import { PolygonFeature } from "./DrawControl";
 import MapControls from "./MapControls";
-import { ADMIN_AREA_CODE, API_BASE_URL } from "../../constants";
+import { ADMIN_AREA_CODE } from "../../constants";
+import { fetchServicesByStops, fetchStops } from "../../data/refDataApi";
 import { PageState } from "../../interfaces";
 import {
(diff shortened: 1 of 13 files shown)


site/components/map/ServicesMap.tsx
```diff
@@ -235,6 +245,26 @@ const Map = ({
         }
     }, [features]);
 
+    useEffect(() => {
+        if (markerData && markerData.length > 0 && services) {
```
lj-kpmg, line 249: I think it would be better if we handle retrieving the service routes when a user clicks select all rather than on drawing the polygon, it's quite an expensive query so we don't want to make it every time the user tweaks a polygon

site/components/map/ServicesMap.tsx
```diff
@@ -13,21 +13,22 @@ import {
     useMemo,
```
lj-kpmg, line 13: I've noticed a couple of issues with the functionality here:

- Hover on unselected stops not showing the tooltip (on stops and services maps)
- Services aren't being added to the list of selected services on clicking select all, stops are added but not services

alainajaffer: The first one I was told to implement it that way for those stops in the polygon that are grey

alainajaffer: fixed the services issue

site/components/map/ServicesMap.tsx
alainajaffer: done

lj-kpmg (approved):

PR #90, opened 2023-05-02
UI changes for login page 

alvak1:
Login page with UI components

### DIFF
--- a/site/pages/login.page.tsx
+++ b/site/pages/login.page.tsx
@@ -1,0 +1,94 @@
+import { ReactElement } from "react";
+import { BaseLayout } from "../components/layout/Layout";
+
+const title = "Sign in - Create Transport Disruptions Service";
+const description = "Login page for the Create Transport Disruptions Service";
+
+const Login = (): ReactElement => (
+    <BaseLayout title={title} description={description} errors={errors}>
+        <div className="govuk-grid-row">
+            <div className="govuk-grid-column-two-thirds">
+                <CsrfForm action="/api/login" method="post" csrfToken={csrfToken}>
+                    <>
+                        <ErrorSummary errors={errors} />
+                        <div className={`govuk-form-group ${errors.length > 0 ? "govuk-form-group--error" : ""}`}>
+                            <fieldset className="govuk-fieldset" aria-describedby="register-page-heading">
+                                <legend className="govuk-fieldset__legend govuk-fieldset__legend--l">
+                                    <h1 className="govuk-fieldset__heading" id="register-page-heading">
+                                        Sign in
+                                    </h1>
+                                </legend>
+                                <p className="govuk-hint hint-text">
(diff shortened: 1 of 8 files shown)


site/schemas/login.schema.ts
```diff
@@ -0,0 +1,9 @@
+import { z } from "zod";
+import { setZodDefaultError } from "../utils";
+
+export const loginSchema = z.object({
+    email: z.string(setZodDefaultError("Enter an email address")).min(1),
+    password: z.string(setZodDefaultError("Enter a password")).min(8),
```
lj-kpmg, line 6: I'd probably just make the min 1 here, if the user accidentally misses a character off their password they might be confused if they get "Enter a password" as the error

alvak1: ive updated the code to use min as 1

site/pages/login.page.tsx
```diff
@@ -0,0 +1,98 @@
...
+    return (
+        <BaseLayout title={title} description={description} errors={pageState.errors}>
+            <h1 className="govuk-heading-xl">Sign in</h1>
+            <div className="govuk-grid-row">
+                <div className="govuk-grid-column-two-thirds">
+                    <CsrfForm action="/api/login" method="post" csrfToken={pageState.csrfToken}>
+                        <>
+                            <ErrorSummary errors={pageState.errors} />
+
+                            <h1 className="govuk-heading-m w-3/4">
```
lj-kpmg, line 33: Make this an `h2`

site/schemas/login.schema.ts
```diff
@@ -0,0 +1,15 @@
+import { z } from "zod";
+import { setZodDefaultError } from "../utils";
+
+export const loginSchema = z.object({
+    email: z.string(setZodDefaultError("Enter an email address")).min(1),
+    password: z
+        .string(setZodDefaultError("Enter a password"))
+        .min(1)
+        .refine((val) => val.length >= 8, {
```
lj-kpmg, line 9: I'd probably just get rid of this refine completely and let cognito do the actual checking of the passwords, this could theoretically lead to some weird behaviour if we changed the password length requirement

site/schemas/login.schema.ts
```diff
@@ -0,0 +1,15 @@
+import { z } from "zod";
+import { setZodDefaultError } from "../utils";
+
+export const loginSchema = z.object({
+    email: z.string(setZodDefaultError("Enter an email address")).min(1),
```
lj-kpmg, line 5: We can use the zod email validation here
```suggestion
    email: z.string().email({ message: "Enter a valid email address" }),
```

site/schemas/login.schema.ts
alvak1: done

site/schemas/login.schema.ts
alvak1: done

PR #95, opened 2023-05-04
change password functionality added

alvak1:
Access the page using /change-password URI


site/constants/index.ts
```diff
@@ -228,6 +230,7 @@ export const COOKIES_CONSEQUENCE_NETWORK_ERRORS = "cdd-consequence-network-error
 export const COOKIES_CONSEQUENCE_STOPS_ERRORS = "cdd-consequence-stops-errors";
 export const COOKIES_CONSEQUENCE_SERVICES_ERRORS = "cdd-consequence-services-errors";
 export const COOKIES_LOGIN_ERRORS = "cdd-consequence-login-errors";
+export const COOKIES_CHANGE_PASSWORD_ERRORS = "cdd-consequence-change-password-errors";
```
RobCattonKPMG, line 233: should this have 'consequence' in the string?

site/pages/api/change-password.api.ts
```diff
@@ -0,0 +1,42 @@
...
+} from "../../utils/apiUtils";
+
+const changePassword = (req: NextApiRequest, res: NextApiResponse) => {
+    try {
+        const validatedBody = changePasswordSchemaRefined.safeParse(req.body);
+        if (!validatedBody.success) {
+            setCookieOnResponseObject(
+                COOKIES_CHANGE_PASSWORD_ERRORS,
+                JSON.stringify({
+                    inputs: req.body as object,
```
RobCattonKPMG, line 19: Do we not want to do any validation on the req.body before we put it in the cookie? If the user edits the form and sends across odd info, we will just put it in the cookie and later on when we try parse it, it will presumably fail?

site/pages/api/change-password.api.ts
```diff
@@ -0,0 +1,42 @@
...
+                    errors: flattenZodErrors(validatedBody.error),
+                }),
+                res,
+            );
+        } else {
+            destroyCookieOnResponseObject(COOKIES_CHANGE_PASSWORD_ERRORS, res);
+        }
+
+        redirectTo(res, `${CHANGE_PASSWORD_PAGE_PATH}${validatedBody.success ? `?success=true` : ""}`);
```
RobCattonKPMG, line 28: `?success=true` doesnt need to be in backticks, it can use normal "" quotes.

site/pages/api/change-password.test.ts
```diff
@@ -0,0 +1,99 @@
...
+        ...(await vi.importActual<object>("../../utils/apiUtils")),
+        setCookieOnResponseObject: vi.fn(),
+        destroyCookieOnResponseObject: vi.fn(),
+    }));
+
+    afterEach(() => {
+        vi.resetAllMocks();
+    });
+
+    it("should redirect to login page when incorrect inputs are passed", () => {
```
RobCattonKPMG, line 20: I think this test name should be 'when no inputs are passed' rather than 'incorrect' ones.

site/pages/change-password.page.tsx
```diff
@@ -0,0 +1,121 @@
...
+    return (
+        <TwoThirdsLayout title={title} description={description} errors={pageState.errors}>
+            <CsrfForm action="/api/change-password" method="post" csrfToken={pageState.csrfToken}>
+                <>
+                    <ErrorSummary errors={pageState.errors} />
+                    <h1 className="govuk-heading-xl">Change password</h1>
+                    {displaySuccessMessage ? (
+                        <>
+                            <h2 className="govuk-heading-m">Your password has been successfully updated.</h2>
+                            <Link role="button" href={ACCOUNT_SETTINGS_PAGE_PATH} className="govuk-button mt-8 ">
```
RobCattonKPMG, line 50:
```suggestion
                            <Link role="button" href={ACCOUNT_SETTINGS_PAGE_PATH} className="govuk-button mt-8">
```

site/pages/change-password.page.tsx
```diff
@@ -0,0 +1,121 @@
...
+                            <TextInput<ChangePasswordProps>
+                                display="Current password"
+                                inputName="currentPassword"
+                                widthClass="w"
+                                value={pageState.inputs.currentPassword}
+                                initialErrors={pageState.errors}
+                                schema={changePasswordSchema.shape.currentPassword}
+                                stateUpdater={stateUpdater}
+                                maxLength={100}
+                                isPassword={true}
```
RobCattonKPMG, line 66: Not a big deal, but we can do isPassword here instead of isPassword={true}. Just a little less text.

site/pages/change-password.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+        confirmPassword: "dummyPassword",
+    },
+};
+
+const withInputsAndErrors: ChangePasswordPageProps = {
+    ...withInputsAndNoErrors,
+    errors: [{ errorMessage: "Enter a valid email address", id: "currentPassword" }],
+};
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
RobCattonKPMG, line 24: We dont need to use requires. If you search the repo for ' vi.spyOn ' you can find examples.

site/schemas/change-password.schema.ts
```diff
@@ -0,0 +1,22 @@
...
+    }),
+});
+
+export type ChangePasswordProps = z.infer<typeof changePasswordSchema>;
+
+export const changePasswordSchemaRefined = changePasswordSchema.refine(
+    (val) => val.newPassword === val.confirmPassword,
+    {
+        path: ["confirmPassword"],
+        message: "You must type the same password each time",
```
RobCattonKPMG, line 20: Unsure if this text has been checked with Steve or anybody, but I think "Both passwords must match" would be a little clearer.

site/pages/api/change-password.api.ts
alvak1: The cookie is set after validation and parsing. The cookie set is an error cookie where the inputs are what is passed back to the user

site/constants/index.ts
alvak1: done

site/pages/api/change-password.api.ts
alvak1: done

site/pages/api/change-password.test.ts
alvak1: done

site/pages/change-password.page.tsx
alvak1: done

site/pages/change-password.page.tsx
alvak1: done

site/pages/login.page.tsx
```diff
@@ -54,6 +54,7 @@ const Login = (props: LoginPageProps): ReactElement => {
                                 schema={loginSchema.shape.password}
                                 stateUpdater={stateUpdater}
                                 maxLength={100}
+                                isPassword={true}
```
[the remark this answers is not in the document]
alvak1: done

site/schemas/change-password.schema.ts
```diff
@@ -0,0 +1,22 @@
+import { z } from "zod";
+import { setZodDefaultError } from "../utils";
+
+export const changePasswordSchema = z.object({
+    currentPassword: z.string(setZodDefaultError("Enter your current password")).min(1),
+    newPassword: z.string().min(8, {
+        message: "Enter a minimum of 8 characters",
+    }),
+    confirmPassword: z.string().min(8, {
```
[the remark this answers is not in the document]
alvak1: done

alvak1 (commented):
submitting review

site/pages/api/change-password.api.ts
RobCattonKPMG: The cookie set uses req.body, which the user sends to the API.

If we are just putting the req.body directly into the cookie, then if the user sends this :

{ 
dinner: "fish"
}

Then we will add it to the cookie, which isnt what we want (presumably it will break when we try and parse the cookie later and dont find what we expect).

site/schemas/change-password.schema.ts
RobCattonKPMG: Ok no worries.

site/pages/api/change-password.api.ts
alvak1: The same schema is used to parse later too - so if using this schema to parse would cause it to break then it would break in the api too before setting the cookie

alvak1: Noticed that the refined schema was not used which is now updated

RobCattonKPMG: Right ok - cool! Looks good then

site/pages/change-password.page.tsx
```diff
@@ -0,0 +1,125 @@
...
+                            <TextInput<ChangePasswordProps>
+                                display="Confirm new password"
+                                inputName="confirmPassword"
+                                widthClass="w"
+                                value={pageState.inputs.confirmPassword}
+                                initialErrors={pageState.errors}
+                                schema={changePasswordSchema.shape.confirmPassword}
+                                stateUpdater={stateUpdater}
+                                maxLength={100}
+                                isPassword={true}
```
lj-kpmg, line 95:
```suggestion
                                isPassword
```

site/schemas/change-password.schema.ts
```diff
@@ -0,0 +1,20 @@
+import { z } from "zod";
+import { setZodDefaultError } from "../utils";
+
+export const changePasswordSchema = z.object({
+    currentPassword: z.string(setZodDefaultError("Enter your current password")).min(1),
+    newPassword: z.string().min(8, {
```
lj-kpmg, line 6: Might be worth making the password length a shared constant since we'll need it on the register page as well

alvak1: Customising error message will not be possible if its used as a shared constant

site/pages/change-password.page.tsx
alvak1: done

site/pages/login.page.tsx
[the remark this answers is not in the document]
alvak1: done

lj-kpmg (approved):

PR #98, opened 2023-05-05
Fix for issue with selecting consequence type page

alvak1:
Navigate to `/type-of-consequence` page and click save without choosing an option. Appropriate error should be displayed

### DIFF
--- a/site/pages/type-of-consequence/[disruptionId]/[consequenceIndex].page.tsx
+++ b/site/pages/type-of-consequence/[disruptionId]/[consequenceIndex].page.tsx
@@ -16,7 +16,7 @@ import {
 import { getDisruptionById } from "../../../data/dynamo";
 import { PageState } from "../../../interfaces/index";
 import { ConsequenceType, typeOfConsequenceSchema } from "../../../schemas/type-of-consequence.schema";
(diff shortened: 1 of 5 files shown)


site/pages/type-of-consequence/[disruptionId]/[consequenceIndex].page.tsx
```diff
@@ -90,6 +90,8 @@ export const getServerSideProps = async (ctx: NextPageContext): Promise<{ props:
         throw new Error("Disruption not found for consequence type page");
     }
 
+    if (ctx.res) destroyCookieOnResponseObject(COOKIES_CONSEQUENCE_TYPE_ERRORS, ctx.res);
```
Comment (lj-kpmg) [inline site/pages/type-of-consequence/[disruptionId]/[consequenceIndex].page.tsx:93]: We don't want to remove the cookie here

Comment (lj-kpmg) [reply site/pages/type-of-consequence/[disruptionId]/[consequenceIndex].page.tsx:93]: In fact I notice this has been added in a few places and will need removing

Comment (lj-kpmg) [reply site/pages/type-of-consequence/[disruptionId]/[consequenceIndex].page.tsx:93]: The errors should persist across a page refresh

lj-kpmg (approved):

PR #101, opened 2023-05-10
feat: Operator filter

RobCattonKPMG:


site/pages/view-all-disruptions.page.tsx
```diff
@@ -403,23 +444,69 @@ const ViewAllDisruptions = ({ disruptions, services, newDisruptionId }: ViewAllD
...
+                        }}
+                    >
+                        Clear filters
+                    </button>
+                </div>
             </div>
 
             {showFilters ? (
                 <>
+                    <OperatorSearch
```
alvak1, line 477: Tried searching for an existing disruption with consequence operator "Arriva Yorkshire" and no data was displayed

site/pages/view-all-disruptions.page.tsx
```diff
@@ -362,6 +379,30 @@ const ViewAllDisruptions = ({ disruptions, services, newDisruptionId }: ViewAllD
...
+                operatorName: op.operatorPublicName,
+                operatorRef: op.nocCode,
+            };
+        });
+        setFilter({
+            ...filter,
+            operators: disruptionOperatorsToSet,
+        });
+        // eslint-disable-next-line react-hooks/rules-of-hooks
+        useFiltersOnDisruptions(
```
alvak1, line 397: Renaming `useFiltersOnDisruptions` to a name without prefix `use` will allow you to remove the eslint disable comment

site/pages/view-all-disruptions.page.tsx
```diff
@@ -362,6 +379,30 @@ const ViewAllDisruptions = ({ disruptions, services, newDisruptionId }: ViewAllD
         ); // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [selectedServices]);
 
+    useEffect(() => {
+        const operatorsToSet = selectedOperatorsNocs.map(
```
alvak1, line 383: We could avoid iterating twice by adding objects to `disruptionOperatorsToSet` while mapping over `selectedOperatorsNocs`

```const disruptionOperatorsToSet: DisruptionOperator[] = [];
        const operatorsToSet = selectedOperatorsNocs.map((selOpNoc) => {
            const operator = operators.find((op) => op.nocCode === selOpNoc) as Operator;
            if (operator) {
                disruptionOperatorsToSet.push({
                    operatorName: operator.operatorPublicName,
                    operatorRef: operator.nocCode,
                });
                return operator;
            }
        });```

File: site/pages/view-all-disruptions.page.tsx (hunk 2 above)
Comment (RobCattonKPMG) [reply site/pages/view-all-disruptions.page.tsx:397]: Ah good catch! Thanks

File: site/pages/view-all-disruptions.page.tsx (hunk 3 above)
Comment (RobCattonKPMG) [reply site/pages/view-all-disruptions.page.tsx:383]: Good idea - the code needed tweaking a little because in a map you need to always return a value, and the if statement was avoiding that. I changed it to a forEach to get around it.

File: site/pages/view-all-disruptions.page.tsx [hunk 4]
```diff
@@ -588,7 +703,7 @@ export const getServerSideProps = async (): Promise<{ props: ViewAllDisruptionsP
                 status: Progress.open,
                 severity: getWorstSeverity(severitys),
                 serviceIds,
-                operators,
+                operators: disruptionOperators,
```
alvak1, line 706: Unable to find where this field is used

site/pages/view-all-disruptions.page.tsx
```diff
@@ -429,112 +516,105 @@ const ViewAllDisruptions = ({ disruptions, services, newDisruptionId }: ViewAllD
 
                     {filter.services.length > 0 ? <Table rows={formatServicesIntoRows(filter, setFilter)} /> : null}
 
-                    {startDateFilterError && (
+                    {(startDateFilterError || endDateFilterError) && (
```
alvak1, line 519: After the data is filtered based on start date and end date - if you delete the end date text then the error should be displayed but its not. 

If you remove both the start date and end date then the table contents should be reverted to the values before start date and end date were applied. This reset is not working.

site/pages/view-all-disruptions.page.tsx
RobCattonKPMG: It is used in the page on line 267 - operators.

site/pages/view-all-disruptions.page.tsx
RobCattonKPMG: Testing this, if I enter a start and end date, and then start deleting the end date, I get an error saying an end date is needed.

RobCattonKPMG: The issue happens when the user drags their cursor over the text and hits delete, as the form doesnt seem to register that the value has been deleted. I'm unsure what we want to do about this, because if we add an onBlur event to check if the input is empty, it will do that even if the user wasn't wanting to enter something.

site/pages/view-all-disruptions.page.tsx
```diff
@@ -403,23 +444,69 @@ const ViewAllDisruptions = ({ disruptions, services, newDisruptionId }: ViewAllD
...
+                        data-module="govuk-button"
+                        onClick={() => {
+                            if (showFilters) {
+                                setShowFilters(false);
+                            } else {
+                                setShowFilters(true);
+                            }
+                        }}
+                    >
+                        {showFilters ? "Hide " : "Show "}filters
```
alvak1, line 459: If filters are added and then hide the filters, followed by clicking show filters - the table displays filtered result but the filter values are not displayed in their respective fields

site/pages/view-all-disruptions.page.tsx
alvak1: the const `operators` from line 659 is used in line 267 as that is what is returned as props in 718 and 728

RobCattonKPMG: Ah ok - the operators: disruptionOperators is what is returned inside the map function, so its the operators which are associated with a disruption.

site/pages/view-all-disruptions.page.tsx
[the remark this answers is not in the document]
RobCattonKPMG: Fixed.

site/pages/view-all-disruptions.page.tsx
RobCattonKPMG: Fixed.

site/pages/view-all-disruptions.page.tsx
alvak1: so is `operators: disruptionOperators` required? If its not then line 720-731 could be redundant?

site/pages/view-all-disruptions.page.tsx
alvak1: Resolving this comment as a new bug will be raised for this issue

site/pages/view-all-disruptions.page.tsx
RobCattonKPMG: yeah its required otherwise we dont know which operators are in a disruption

alvak1 (dismissed):
Approved

PR #104, opened 2023-05-15
feat: Admin remove

alainajaffer:
Go to the user management page after making a user and then delete that user by clicking 'remove'


site/pages/api/delete-user.api.ts
```diff
@@ -0,0 +1,20 @@
+import { NextApiResponse } from "next";
+import { deleteAdminUser } from "../../data/cognito";
+import { DeleteUserApiRequest } from "../../interfaces";
+import { redirectTo, redirectToError } from "../../utils/apiUtils";
+
+const deleteUser = async (req: DeleteUserApiRequest, res: NextApiResponse): Promise<void> => {
```
lj-kpmg, line 6: We need to restrict access to this api endpoint so only authenticated admins can call it, if we update line 173 in `middleware.api.ts` to:

```ts
if (request.nextUrl.pathname.startsWith("/admin/") || request.nextUrl.pathname.startsWith("/api/admin/")) {
```

and move this file into an admin folder then that should do the job, the action endpoint for the form will need updating too

site/pages/api/delete-user.api.ts
```diff
@@ -0,0 +1,20 @@
...
+    try {
+        const { username } = req.body;
+
+        await deleteAdminUser(username);
```
lj-kpmg, line 10: Thinking about it we probably need to be more defensive here, we should verify that the user being deleted has the same orgId as the admin doing the deleting, we can grab the logged in user's orgId by calling `getSession` and then we probably need a new function in cognito.ts to get a user's details by their username using `AdminGetUserCommand`

site/data/cognito.ts
```diff
@@ -30,6 +33,22 @@ const cognito = new CognitoIdentityProviderClient({
     region: "eu-west-2",
 });
 
+export const deleteAdminUser = (username: string) => {
```
lj-kpmg, line 36: I would just call this deleteUser

site/interfaces/index.ts
```diff
@@ -28,6 +28,12 @@ export interface CookiesApiRequest extends NextApiRequest {
     };
 }
 
+export interface DeleteUserApiRequest extends NextApiRequest {
```
lj-kpmg, line 31: I would just declare this in the api file if it's only used there

site/pages/api/delete-user.api.ts
alainajaffer: done

site/pages/api/delete-user.api.ts
alainajaffer: done

site/data/cognito.ts
alainajaffer: done

site/interfaces/index.ts
alainajaffer: done

site/pages/api/admin/delete-user.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    vi.mock("../../../data/cognito", () => ({
+        deleteUser: vi.fn(),
+        getUserDetails: vi.fn(),
+    }));
+
+    afterEach(() => {
+        vi.resetAllMocks();
+    });
+
+    it("should redirect to /admin/user-management if delete was a success", async () => {
```
lj-kpmg, line 23: Can we have a test covering the case where the org ids don't match?

site/data/cognito.ts
```diff
@@ -30,6 +34,38 @@ const cognito = new CognitoIdentityProviderClient({
...
+            throw new Error(`Failed to delete user: ${error.stack || ""}`);
+        }
+
+        throw error;
+    }
+};
+
+export const getUserDetails = (username: string) => {
+    try {
+        const params: AdminDeleteUserRequest = {
```
lj-kpmg, line 55:
```suggestion
        const params: AdminGetUserCommandInput = {
```

site/data/cognito.ts
```diff
@@ -30,6 +34,38 @@ const cognito = new CognitoIdentityProviderClient({
     region: "eu-west-2",
 });
 
+export const deleteUser = (username: string) => {
+    try {
+        const params: AdminDeleteUserRequest = {
```
lj-kpmg, line 39:
```suggestion
        const params: AdminDeleteUserCommandInput = {
```

site/pages/api/admin/delete-user.test.ts
alainajaffer: done

site/data/cognito.ts
alainajaffer: done

site/data/cognito.ts
alainajaffer: done

site/pages/admin/user-management.page.tsx
```diff
@@ -49,21 +51,51 @@ const UserManagement = ({ userList }: UserManagementPageProps): ReactElement =>
...
             <>
-                <Link key={`${key}${index ? `-${index}` : ""}`} className="govuk-link" href="/">
-                    Resend invite
-                </Link>
-                <br />
-                <Link key={`${key}${index ? `-remove-${index}` : "-remove"}`} className="govuk-link" href="/">
-                    Remove
-                </Link>
+                {showRemovePopup ? (
+                    <DeleteConfirmationPopup
```
lj-kpmg, line 66: This is creating a modal for each link so it will actually be rendering multiple modals on top of each other and the top one will always be for the last user in the list, the modal should be moved out to the top level and the username to be deleted will need to be set in state, something like:

```ts
const UserManagement = ({ userList, csrfToken }: UserManagementPageProps): ReactElement => {
    const numberOfUserPages = Math.ceil(userList.length / 10);
    const [currentPage, setCurrentPage] = useState(1);
    const [userToDelete, setUserToDelete] = useState<string | null>(null);

    const getAccountType = (groupName: UserGroups): string => {
        switch (groupName) {
            case UserGroups.systemAdmins:
            case UserGroups.orgAdmins:
                return "Admin";
            case UserGroups.orgPublishers:
                return "Publisher";
            case UserGroups.orgStaff:
                return "Staff";
        }
    };

    const getRows = () => {
        const rows: { header?: string | ReactNode; cells: string[] | ReactNode[] }[] = [];
        getDataInPages(currentPage, userList).forEach((user, index) => {
            rows.push({
                cells: [
                    `${getAccountType(user.group)}`,
                    user.email,
                    user.userStatus === "CONFIRMED" ? "Active" : "Pending invite",
                    createLink("user-action", index, user.username, user.userStatus === "CONFIRMED" ? false : true),
                ],
            });
        });
        return rows;
    };

    const removeUser = (username: string) => {
        setUserToDelete(username);
    };

    const cancelActionHandler = () => {
        setUserToDelete(null);
    };

    const createLink = (key: string, index: number, username: string, sendInvite?: boolean) => {
        return (
            <>
                {sendInvite ? (
                    <>
                        <Link key={`${key}${index ? `-${index}` : ""}`} className="govuk-link" href="/">
                            Resend invite
                        </Link>
                        <br />
                        <button
                            key={`${key}${index ? `-remove-${index}` : "-remove"}`}
                            className="govuk-link"
                            onClick={() => removeUser(username)}
                        >
                            Remove
                        </button>
                    </>
                ) : (
                    <button
                        key={`${key}${index ? `-${index}` : ""}`}
                        className="govuk-link"
                        onClick={() => removeUser(username)}
                    >
                        Remove
                    </button>
                )}
            </>
        );
    };

    return (
        <BaseLayout title={title} description={description}>
            <>
                {userToDelete ? (
                    <DeleteConfirmationPopup
                        entityName="user"
                        deleteUrl="/api/admin/delete-user"
                        cancelActionHandler={cancelActionHandler}
                        csrfToken={csrfToken || ""}
                        hiddenInputs={[
                            {
                                name: "username",
                                value: userToDelete,
                            },
                        ]}
                    />
                ) : null}
                    <h1 className="govuk-heading-xl">User Management</h1>
                </div>
                <Table
                    caption={{ text: "Users", size: "m" }}
                    columns={["Account type", "User email", "Status", "Action"]}
                    rows={getRows()}
                ></Table>
                <Link role="button" href={"/admin/add-user"} className="govuk-button--secondary govuk-button mt-5">
                    Add new user
                </Link>
                <PageNumbers
                    numberOfPages={numberOfUserPages}
                    currentPage={currentPage}
                    setCurrentPage={setCurrentPage}
                />
            </>
        </BaseLayout>
    );
};
```

alainajaffer: done

site/pages/api/admin/delete-user.api.ts
```diff
@@ -0,0 +1,41 @@
...
+        username: string;
+    };
+}
+
+const deleteUser = async (req: DeleteUserApiRequest, res: NextApiResponse): Promise<void> => {
+    try {
+        const { username } = req.body;
+
+        const session = getSession(req);
+        if ((session && !session.orgId) || !session) {
```
RobCattonKPMG, line 19: could maybe check for username here also

PR #105, opened 2023-05-15
Add cognito users

alvak1:
Access the page `/admin/user-management` and then click on add user button


site/pages/api/add-user.api.ts
```diff
@@ -0,0 +1,67 @@
...
+import { addUserSchema } from "../../schemas/add-user.schema";
+import { flattenZodErrors } from "../../utils";
+import {
+    redirectToError,
+    setCookieOnResponseObject,
+    redirectTo,
+    destroyCookieOnResponseObject,
+} from "../../utils/apiUtils";
+
+const addUser = async (req: NextApiRequest, res: NextApiResponse) => {
```
lj-kpmg, line 14: I've mentioned this in Alaina's PR but I'll put it here as well: We need to restrict access to this api endpoint so only authenticated admins can call it, if we update line 173 in middleware.api.ts to:
```ts
if (request.nextUrl.pathname.startsWith("/admin/") || request.nextUrl.pathname.startsWith("/api/admin/")) {
```
and move this file into an admin folder then that should do the job, the action endpoint for the form will need updating too

site/pages/api/add-user.api.ts
```diff
@@ -0,0 +1,67 @@
...
+                    errors: flattenZodErrors(validatedBody.error),
+                }),
+                res,
+            );
+
+            redirectTo(res, ADD_USER_PAGE_PATH);
+            return;
+        }
+
+        await createUser(validatedBody.data);
```
lj-kpmg, line 31: We should get the session of the logged in user and make sure that the user they're trying to create is in the same org, otherwise admin users could create other users in any organisation

site/pages/admin/user-management.page.tsx
```diff
@@ -113,7 +113,13 @@ export const getServerSideProps = async (ctx: NextPageContext): Promise<{ props:
 
     const userList = userManagementSchema
         .parse(userRecords)
-        .filter((user) => user.organisation === sessionWithOrg.orgId);
+        .filter((user) => user.organisation === sessionWithOrg.orgId)
+        .sort((a, b) => {
+            return (
+                a.givenName.toLowerCase().localeCompare(b.givenName.toLowerCase()) ||
```
lj-kpmg, line 119: I would just sort by email here since we don't display first and last names on the page

site/pages/admin/add-user.page.tsx
```diff
@@ -0,0 +1,131 @@
...
+                        widthClass="w"
+                        value={pageState.inputs.email}
+                        initialErrors={pageState.errors}
+                        schema={addUserSchema.shape.email}
+                        stateUpdater={stateUpdater}
+                        maxLength={100}
+                    />
+
+                    <Table rows={[{ header: "Organisation", cells: [pageState.sessionWithOrg?.orgName, ""] }]} />
+                    <input type="hidden" name="orgId" value={pageState.sessionWithOrg?.orgId} />
```
lj-kpmg, line 67: This should be retrieved from the session directly in the API call, otherwise people could pass through whatever orgId they want

site/pages/api/add-user.api.ts
alvak1: done

site/pages/api/add-user.api.ts
alvak1: done. There is no input for org id. The id is retrieved from the session within the api and passed to cognito call.

site/pages/admin/user-management.page.tsx
alvak1: done

site/pages/admin/add-user.page.tsx
alvak1: done. More details in comment above

stacks/CognitoStack.ts
```diff
@@ -28,6 +28,7 @@ export const CognitoStack = ({ stack }: StackContext) => {
         timeout: 30,
         memorySize: 256,
         runtime: "python3.9",
+        enableLiveDev: false,
```
alainajaffer, line 31: why is this required?

lj-kpmg (approved):

stacks/CognitoStack.ts
alvak1: The admin create user command was timing out and expected the stack to be up locally

PR #106, opened 2023-05-16
Restricted user to publish disruptions using pending approval status

alvak1:
Testing instructions:

Click create new disruptions in dashboard page and proceed upto Review Disruption page. Restricted users should be displayed "Send to review" button instead of "Publish" button. The same is applicable when disruption is edited in disruption detail page


site/data/dynamo.ts
```diff
@@ -67,6 +66,29 @@ export const getPublishedDisruptionsDataFromDynamo = async (id: string): Promise
     return disruptionIds?.map((id) => collectDisruptionsData(dbData.Items || [], id)).filter(notEmpty) ?? [];
 };
 
+export const getSubmittedDisruptionsDataFromDynamo = async (id: string): Promise<Disruption[]> => {
+    logger.info("Getting disruptions data from DynamoDB table...");
```
RobCattonKPMG, line 70: Should we change the log here and the one in getPublishedDisruptionsDataFromDynamo so that they're different? As they're doing different things.

site/data/dynamo.ts
```diff
@@ -225,7 +248,9 @@ export const upsertDisruptionInfo = async (disruptionInfo: DisruptionInfo, id: s
     logger.info(`Updating draft disruption (${disruptionInfo.disruptionId}) in DynamoDB table...`);
     const currentDisruption = await getDisruptionById(disruptionInfo.disruptionId, id);
     const isEditing =
-        currentDisruption?.publishStatus === "PUBLISHED" || currentDisruption?.publishStatus === "EDITING";
+        currentDisruption?.publishStatus === "PUBLISHED" ||
```
RobCattonKPMG, line 251: I wonder if we should have a publish status enum, as these values seem like good candidates for an enum.

site/data/dynamo.ts
```diff
@@ -67,6 +66,29 @@ export const getPublishedDisruptionsDataFromDynamo = async (id: string): Promise
     return disruptionIds?.map((id) => collectDisruptionsData(dbData.Items || [], id)).filter(notEmpty) ?? [];
 };
 
+export const getSubmittedDisruptionsDataFromDynamo = async (id: string): Promise<Disruption[]> => {
```
lj-kpmg, line 69: We don't want to write to the siri table in the case that the disruption is pending approval

site/data/dynamo.ts
```diff
@@ -164,6 +186,7 @@ export const insertPublishedDisruptionIntoDynamoAndUpdateDraft = async (
     ptSituationElement: PtSituationElement,
     disruption: Disruption,
     id: string,
+    status: string,
```
lj-kpmg, line 189: This function will also need updating so that we don't write to the siri table in the case that the disruption is pending approval

site/data/dynamo.ts
alvak1: done

site/data/dynamo.ts
alvak1: done

site/pages/review-disruption/review-disruption.test.tsx
```diff
@@ -105,6 +117,21 @@ describe("pages", () => {
...
+        });
+
+        it("should render correctly with inputs and staff user role", () => {
+            const tree = renderer
+                .create(
+                    <ReviewDisruption
+                        disruption={previousDisruptionInformation}
+                        previousSocialMediaPosts={previousSocialMediaPosts}
+                        errors={[]}
+                        session={{ ...userSession, isSystemAdmin: false, isOrgStaff: true }}
```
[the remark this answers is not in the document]
alvak1: done

site/pages/disruption-detail/disruption-detail.test.tsx
```diff
@@ -101,6 +118,24 @@ describe("pages", () => {
...
+
+        it("should render correctly with inputs and user role set to staff", () => {
+            const tree = renderer
+                .create(
+                    <DisruptionDetail
+                        disruption={{
+                            ...previousDisruptionInformation,
+                            publishStatus: "EDITING",
+                        }}
+                        redirect={"/view-all-disruptions"}
```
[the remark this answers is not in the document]
alvak1: done

site/data/dynamo.ts
alvak1: done

site/schemas/disruption.schema.ts
```diff
@@ -5,7 +5,9 @@ import { createDisruptionsSchemaRefined } from "./create-disruption.schema";
 export const disruptionSchema = createDisruptionsSchemaRefined.and(
     z.object({
         consequences: z.array(consequenceSchema).optional(),
-        publishStatus: z.union([z.literal("DRAFT"), z.literal("PUBLISHED"), z.literal("EDITING")]).default("DRAFT"),
+        publishStatus: z
+            .union([z.literal("DRAFT"), z.literal("PUBLISHED"), z.literal("EDITING"), z.literal("PENDING APPROVAL")])
```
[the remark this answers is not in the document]
alvak1: done

site/data/dynamo.ts
```diff
@@ -164,6 +187,7 @@ export const insertPublishedDisruptionIntoDynamoAndUpdateDraft = async (
...
```
lj-kpmg, line 190: This can be of type PublishStatus

site/data/dynamo.ts
```diff
@@ -184,24 +208,31 @@ export const insertPublishedDisruptionIntoDynamoAndUpdateDraft = async (
...
                 UpdateExpression: "SET publishStatus = :1",
                 ExpressionAttributeValues: {
-                    ":1": "PUBLISHED",
+                    ":1": status,
                 },
             },
         })) ?? [];
 
+    const putSiriTable =
+        status !== PublishStatus.pendingApproval
```
lj-kpmg, line 217: Has the change to the enums file been pushed up?

lj-kpmg, line 217: I would make this `status === PublishStatus.published` in case we extend this to handle more statuses in the future since we only want to push to the siri in the case where the disruption is published

alvak1: done

lj-kpmg (changes requested):
I've noticed that when editing a published disruption as a staff member, it sets the status to pending approval but it overwrites the original disruption with the changes, so if the change is rejected we won't have the original disruption to revert to, we'll need to keep the edit items in the db until the change has been approved

PR #108, opened 2023-05-16
feat: Admin resend invite

alainajaffer:
On the user-management.page.tsx change line 80 ` {sendInvite ? (` to ` {true ? (` then click on 'Resend' click 'yes, resend' (next to the user on the browser) you should receive an email and the user should be deleted from the table


site/pages/api/admin/resend-invite.api.ts
```diff
@@ -0,0 +1,48 @@
...
+
+        const userDetails = await getUserDetails(username);
+
+        const validatedBody = user.safeParse({ ...userDetails, group });
+
+        if (!validatedBody.success) {
+            throw Error("Insufficient values provided to resend an invite");
+        }
+
+        if (validatedBody.data.orgId !== session.orgId) {
```
lj-kpmg, line 32: We should also verify that the user is in a pending status so that we don't delete confirmed users, the status should be "FORCE_CHANGE_PASSWORD"

alainajaffer: done

lj-kpmg (approved):

PR #121, opened 2023-05-23
feat: Added new type of disruption status and mapped value to a display value

RobCattonKPMG:


site/pages/view-all-disruptions.page.tsx
```diff
@@ -87,27 +87,31 @@ export interface Filter {
...
 
-export const getDisruptionStatus = (disruption: SortedDisruption): Progress => {
-    if (disruption.publishStatus === "DRAFT") {
-        return Progress.draft;
+const getDisruptionStatus = (disruption: SortedDisruption): string => {
+    if (disruption.publishStatus === PublishStatus.draft) {
+        return "draft";
+    }
+
+    if (disruption.publishStatus === PublishStatus.pendingApproval) {
```
alvak1, line 95: Should this be add as an option in the Status filter?

PR #131, opened 2023-05-31
fix: Image lost on page refresh error

alainajaffer:
Show an error message when an error is triggered on the page and the user had uploaded an image before. It will prompt them to upload an image again


alvak1 (approved):
Approved

PR #149, opened 2023-06-27
fix: Remove change link on successful social media post

alainajaffer:
And if disruption is rejected but then editing it becomes pending so you can publish again
And on test the time for a post is scheduled one hour ahead of dev


alvak1 (approved):
Approved

PR #151, opened 2023-06-28
fix: When deleting yourself sign out

alainajaffer:
Go to user management and delete yourself - you should be logged out and your user deleted - so when you try to login you should be unable to


alvak1 (approved):
Approved

PR #152, opened 2023-06-28
Allow org admins to toggle source data between tnds and bods

alvak1:
Testing instructions

- Login as org admin user and then access the Account Settings page from top right corner. The radio buttons should be displayed only to org admins
- After the source data is clicked - navigate to consequence services page and observe that the services data changes based on the source for mode of transport
- The services dropdown in view  all disruptions page should display both tnds and bods data

### DIFF
--- a/shared-ts/enums.ts
+++ b/shared-ts/enums.ts
@@ -176,3 +176,8 @@ export enum SocialMediaPostStatus {
     successful = "Successful",
     rejected = "Rejected",
 }
+
+export enum Modes {
+    tnds = "tnds",
+    bods = "bods",
+}
--- a/site/components/map/ServicesMap.tsx
+++ b/site/components/map/ServicesMap.tsx
@@ -29,7 +29,11 @@ import {
 } from "../../schemas/consequence.schema";
 import { flattenZodErrors } from "../../utils";
 import { getStopType, sortStops } from "../../utils/formUtils";
+import { Modes } from "@create-disruptions-data/shared-ts/enums";
 
+interface ServiceMapProps extends MapProps {
+    dataSource?: Modes;
+}
 interface MapProps {
     initialViewState: Partial<ViewState>;
     style: CSSProperties;
@@ -41,7 +45,7 @@ interface MapProps {
     stateUpdater: Dispatch<SetStateAction<PageState<Partial<ServicesConsequence>>>>;
     state: PageState<Partial<ServicesConsequence>>;
     searchedRoutes?: Partial<(Routes & { serviceId: number })[]>;
-    services?: Service[];
+    services?: Service[] | null;
 }
 
 const lineLayout: LineLayout = {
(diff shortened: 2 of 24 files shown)


site/components/map/ServicesMap.tsx
```diff
@@ -41,7 +45,7 @@ interface MapProps {
     stateUpdater: Dispatch<SetStateAction<PageState<Partial<ServicesConsequence>>>>;
     state: PageState<Partial<ServicesConsequence>>;
     searchedRoutes?: Partial<(Routes & { serviceId: number })[]>;
-    services?: Service[];
+    services?: Service[] | null;
```
alainajaffer, line 48: why was this added?

site/interfaces/index.ts
```diff
@@ -68,7 +68,8 @@ export interface PageState<T> {
 }
 
 export interface CreateConsequenceProps {
-    initialServices?: Service[];
+    initialBodsServices?: Service[] | null;
```
alainajaffer, line 71: why is this | null can it not just be optional and  Service[]?

site/utils/formUtils.ts
```diff
@@ -84,3 +85,29 @@ export const getStopType = (stopType: string | undefined) => {
         return "Stop";
     }
 };
+
+// eslint-disable-next-line @typescript-eslint/require-await
+export const filterServices = async (servicesData: Service[] | undefined): Promise<Service[] | null> => {
+    return new Promise((resolve) => {
```
alainajaffer, line 91: why return a promise here?

site/utils/formUtils.ts
```diff
@@ -84,3 +85,29 @@ export const getStopType = (stopType: string | undefined) => {
...
+        let services: Service[] | null = null;
```
alainajaffer, line 92: why not make this an empty array?

site/utils/formUtils.ts
```diff
@@ -84,3 +85,29 @@ export const getStopType = (stopType: string | undefined) => {
         return "Stop";
     }
 };
+
+// eslint-disable-next-line @typescript-eslint/require-await
+export const filterServices = async (servicesData: Service[] | undefined): Promise<Service[] | null> => {
```
alainajaffer, line 90: why not make this an optional parameter instead of | undefined

site/components/map/ServicesMap.tsx
alvak1: done - removed null

site/utils/formUtils.ts
alvak1: done - removed promise as its implicitly added

site/utils/formUtils.ts
alvak1: done - updated to empty array

shared-ts/enums.ts
```diff
@@ -176,3 +176,8 @@ export enum SocialMediaPostStatus {
     successful = "Successful",
     rejected = "Rejected",
 }
+
```
alainajaffer, line 179: sometimes when i change a mode of transport the services / stops clear - is this expected?

site/utils/formUtils.ts
```diff
@@ -84,3 +85,27 @@ export const getStopType = (stopType: string | undefined) => {
         return "Stop";
     }
 };
+
+// eslint-disable-next-line @typescript-eslint/require-await
+export const filterServices = async (servicesData?: Service[]) => {
```
alainajaffer, line 90: can you remove the async

alvak1: If async was removed then the method would run synchronously and not in parallel

alainajaffer: I don't think it will make too much of a difference to performance in this case

alainajaffer (approved):

PR #153, opened 2023-06-29
fix: fixed duplicate connection issue and added try catch to understand error

alainajaffer:
try connect to the hootsuite account twice it will throw an error second time

added try catch for the refresh token issue

### DIFF
--- a/site/data/s3.ts
+++ b/site/data/s3.ts
@@ -54,3 +54,27 @@ export const getItem = async (bucket: string, key: string, originalFilename: str
         throw error;
     }
 };
+
+export const getObject = async (bucket: string, key: string, originalFilename: string): Promise<Uint8Array | null> => {
+    logger.info("", {
+        context: "data.s3",
+        message: "getting item from s3",
+    });
+
+    try {
+        const input = {
+            Bucket: bucket,
+            Key: key,
+            ResponseContentDisposition: 'attachment; filename ="' + originalFilename + '"',
+        };
+        const command = new GetObjectCommand(input);
+        const response = await s3.send(command);
+        return response.Body?.transformToByteArray() ?? null;
+    } catch (error) {
+        if (error instanceof Error) {
+            throw new Error(`Failed to get item from s3: ${error.stack || ""}`);
+        }
+
+        throw error;
+    }
+};
--- a/site/pages/api/hootsuite-callback.api.ts
+++ b/site/pages/api/hootsuite-callback.api.ts
@@ -101,9 +101,7 @@ const hootsuiteCallback = async (req: NextApiRequest, res: NextApiResponse) => {
 
         const key = `/social/${session.orgId}/hootsuite/${userId}-token`;
 
-        const addedByKey = `/social/${session.orgId}/hootsuite/${userId}-addedUser-${
-            session.name?.replace(" ", "_") || session.username
-        }`;
+        const addedByKey = `/social/${session.orgId}/hootsuite/${userId}-addedUser`;
         await putParameter(key, tokenResult.refresh_token, "SecureString", true);
         await putParameter(addedByKey, session.name?.replace(" ", "_") || session.username, "SecureString", true);
         redirectTo(res, SOCIAL_MEDIA_ACCOUNTS_PAGE_PATH);
--- a/site/pages/api/hootsuite-callback.api.ts
+++ b/site/pages/api/hootsuite-callback.api.ts
@@ -113,6 +113,8 @@ const hootsuiteCallback = async (req: NextApiRequest, res: NextApiResponse) => {
                 ),
             );
         }
+
+        // delay added to make sure delete actions complete before redirection
         await delay(3000);
 
         const key = `/social/${session.orgId}/hootsuite/${userId}-${
--- a/site/pages/api/hootsuite-callback.api.ts
+++ b/site/pages/api/hootsuite-callback.api.ts
@@ -11,7 +11,7 @@ import { getParameter, putParameter } from "../../data/ssm";
 import { initiateRefreshAuth } from "../../middleware.api";
 import { hootsuiteMeSchema, hootsuiteTokenSchema } from "../../schemas/hootsuite.schema";
 import { sessionSchema } from "../../schemas/session.schema";
-import { redirectToError, redirectTo, setCookieOnResponseObject, delay } from "../../utils/apiUtils/index";
+import { redirectToError, redirectTo, setCookieOnResponseObject } from "../../utils/apiUtils/index";
 
 const hootsuiteCallback = async (req: NextApiRequest, res: NextApiResponse) => {
     try {
--- a/site/pages/api/hootsuite-callback.api.ts
+++ b/site/pages/api/hootsuite-callback.api.ts
@@ -7,7 +7,7 @@ import {
     HOOTSUITE_URL,
     SOCIAL_MEDIA_ACCOUNTS_PAGE_PATH,
 } from "../../constants";
-import { deleteParameter, getParameter, getParametersByPath, putParameter } from "../../data/ssm";
+import { getParameter, putParameter } from "../../data/ssm";
 import { initiateRefreshAuth } from "../../middleware.api";
 import { hootsuiteMeSchema, hootsuiteTokenSchema } from "../../schemas/hootsuite.schema";
 import { sessionSchema } from "../../schemas/session.schema";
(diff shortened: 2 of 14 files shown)


site/pages/api/hootsuite-callback.api.ts
```diff
@@ -101,10 +101,21 @@ const hootsuiteCallback = async (req: NextApiRequest, res: NextApiResponse) => {
...
 
+        let refreshTokens = null;
+        try {
+            refreshTokens = (await getParametersByPath(`/social/${session.orgId}/hootsuite`)) ?? null;
+        } catch (e) {}
+
+        const refreshTokenExists = refreshTokens?.Parameters?.find((rt) => rt.Name?.includes(userId));
+
+        if (refreshTokenExists) {
+            throw new Error("Refresh token already exists");
```
alvak1, line 112: Why are we throwing an error?

alvak1: If we have received a new token from Hootsuite then this should be updated in SSM - if we use an older refresh token then the api's will no longer work

alainajaffer: this is if a different use hits connect then another token is made as they become the creator of a new refresh token - it does not overwrite. The connection should only happen once to hootsuite

alvak1: if the refresh token for this other user was already available and the user has clicked the connect to hootsuite button again - in this case an error would be thrown but the refresh token would be lost

alainajaffer: it shouldn't be available in the first place there should only be one

alvak1: as discussed the data displayed in the table should be for a user but the latest refresh token should also be saved in SSM

alainajaffer: done

alainajaffer:
Note: Added a delay because it was taking time to run the delete parameter ssm actions hence producing duplicate entries. We may need to move this to a db approach as tech debt to make this efficient.

site/pages/api/hootsuite-callback.api.ts
```diff
@@ -101,10 +99,27 @@ const hootsuiteCallback = async (req: NextApiRequest, res: NextApiResponse) => {
...
+        const refreshTokenExists = refreshTokens?.Parameters?.filter((rt) => rt.Name?.includes(userId));
+
+        if (refreshTokenExists && refreshTokenExists.length > 0) {
+            await Promise.all(
+                refreshTokenExists.map(async (refreshToken) =>
+                    refreshToken.Name ? await deleteParameter(refreshToken.Name) : "",
+                ),
+            );
+        }
+        await delay(3000);
```
alvak1, line 116: Can we add comment to indicate there is eventual consistency is SSM for deleteParameter api

alainajaffer: done

site/data/s3.ts
```diff
@@ -54,3 +54,27 @@ export const getItem = async (bucket: string, key: string, originalFilename: str
...
+
+    try {
+        const input = {
+            Bucket: bucket,
+            Key: key,
+            ResponseContentDisposition: 'attachment; filename ="' + originalFilename + '"',
+        };
+        const command = new GetObjectCommand(input);
+        const response = await s3.send(command);
+        return response.Body?.transformToByteArray() ?? null;
```
alvak1, line 72: Could we return response.Body and so that the method `transformToByteArray` is invoked at the client end?

site/pages/api/hootsuite-callback.api.ts
```diff
@@ -101,10 +99,13 @@ const hootsuiteCallback = async (req: NextApiRequest, res: NextApiResponse) => {
 
         const userId: string = userDetails.data.id;
 
-        const key = `/social/${session.orgId}/hootsuite/${userId}-${
+        const key = `/social/${session.orgId}/hootsuite/${userId}-token`;
+
+        const addedByKey = `/social/${session.orgId}/hootsuite/${userId}-addedUser-${
```
alvak1, line 104: Do we need the prefix after addedUser? If another user logs in then are we invoking the `deleteParameter` api for this user?

site/pages/api/publish-edit.api.ts
```diff
@@ -33,6 +33,7 @@ const publishEdit = async (req: NextApiRequest, res: NextApiResponse) => {
             redirectTo(res, ERROR_PATH);
             return;
         }
+        logger.info("Retrieved session");
```
alvak1, line 36: Could this be updated to logger.debug

site/pages/api/publish-edit.api.ts
```diff
@@ -56,6 +58,7 @@ const publishEdit = async (req: NextApiRequest, res: NextApiResponse) => {
             return;
         }
 
+        logger.info("Validated zod body");
```
alvak1, line 61: logger.debug

site/pages/api/publish-edit.api.ts
```diff
@@ -90,15 +93,23 @@ const publishEdit = async (req: NextApiRequest, res: NextApiResponse) => {
...
+            await publishToHootsuite(
+                validatedDisruptionBody.data.socialMediaPosts,
+                session.orgId,
+                session.isOrgStaff,
+                canPublish(session),
+            );
         }
-        cleardownCookies(req, res);
 
+        logger.info("Published social media");
```
alvak1, line 110: Could we delete this log

site/pages/api/publish-edit.api.ts
```diff
@@ -90,15 +93,23 @@ const publishEdit = async (req: NextApiRequest, res: NextApiResponse) => {
                   session.name,
               );
 
+        logger.info("Disruption related actions completed");
```
alvak1, line 96: logger.debug

site/pages/dashboard.page.tsx
```diff
@@ -286,6 +287,7 @@ export const getServerSideProps = async (ctx: NextPageContext): Promise<{ props:
         return baseProps;
     }
 
+    logger.info("here on the dashboard");
```
alvak1, line 290: Could we delete this

site/pages/dashboard.page.tsx
```diff
@@ -13,6 +13,7 @@ import { Disruption } from "../schemas/disruption.schema";
 import { getSortedDisruptionFinalEndDate, reduceStringWithEllipsis, sortDisruptionsByStartDate } from "../utils";
 import { canPublish, getSessionWithOrgDetail } from "../utils/apiUtils/auth";
 import { convertDateTimeToFormat, getDate, getDatetimeFromDateAndTime, isLiveDisruption } from "../utils/dates";
+import logger from "../utils/logger";
```
alvak1, line 16: Could we delete this

site/pages/api/publish-edit.api.ts
```diff
@@ -43,6 +44,7 @@ const publishEdit = async (req: NextApiRequest, res: NextApiResponse) => {
             return;
         }
 
+        logger.info("Retrieved draft disruption");
```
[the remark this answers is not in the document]
alainajaffer: decided to remove this

site/pages/api/publish-edit.api.ts
alainajaffer: decided to remove this

site/pages/api/publish-edit.api.ts
```diff
@@ -90,15 +93,23 @@ const publishEdit = async (req: NextApiRequest, res: NextApiResponse) => {
...
+                session.orgId,
+                session.isOrgStaff,
+                canPublish(session),
+            );
         }
-        cleardownCookies(req, res);
 
+        logger.info("Published social media");
+        cleardownCookies(req, res);
+        logger.info("Cleared cookies");
```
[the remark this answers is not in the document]
alainajaffer: done

site/pages/dashboard.page.tsx
alainajaffer: done

site/pages/dashboard.page.tsx
alainajaffer: done

site/pages/api/hootsuite-callback.api.ts
alainajaffer: yes for clarity is someone checks ssm to understand what that value is and it will get overwritten when another user connects

site/pages/api/publish-edit.api.ts
alainajaffer: removed this

site/data/s3.ts
alvak1: If in future we need to read files from S3 then we'll have to make changes to tested code

site/pages/api/hootsuite-callback.api.ts
alvak1: As parameter value will contain the user name - this could be used to check the value.  

For overwriting, we are calling deleteParameter and then putParameter - If this is done to overcome eventual consistency in SSM then its fine but if we are not facing the issue while updating the existing parameter then we can avoid the redundant deleteParamter api call

site/data/s3.ts
alainajaffer: as discussed

site/pages/api/hootsuite-callback.api.ts
alainajaffer: done

alvak1 (approved):
Approve

PR #159, opened 2023-07-05
feat: Duplicate Consequence

alainajaffer:


site/pages/api/duplicate-consequence.api.ts
```diff
@@ -0,0 +1,65 @@
...
+        const consequenceToDuplicate = disruption.consequences.find(
+            (consequence) => consequence.consequenceIndex === Number(consequenceId),
+        );
+
+        if (!consequenceToDuplicate) {
+            throw new Error("No consequence found to duplicate");
+        }
+
+        await upsertConsequence(
+            { ...consequenceToDuplicate, consequenceIndex: disruption.consequences.length },
```
alvak1, line 46: The consequenceIndex should be 1 plus the largest index in consequences array. Use of length does not consider deleted consequences. 

For example Initially they could be 3 consequences added and then the first consequence is deleted. Using length would mean that it would update the last consequence instead of creating a new one.

alvak1 (dismissed):
Approved

site/pages/api/duplicate-consequence.api.ts
alainajaffer: done

site/pages/api/duplicate-consequence.api.ts
```diff
@@ -43,7 +43,15 @@ const duplicateConsequence = async (req: NextApiRequest, res: NextApiResponse):
         }
 
         await upsertConsequence(
-            { ...consequenceToDuplicate, consequenceIndex: disruption.consequences.length },
+            {
+                ...consequenceToDuplicate,
+                consequenceIndex: disruption.consequences
+                    ? disruption.consequences.reduce(
+                          (max, c) => (c.consequenceIndex > max ? c.consequenceIndex : max),
+                          disruption.consequences[0].consequenceIndex,
```
alvak1, line 51: Could this be 0

alainajaffer: no because let's say i delete them all and i have a consequence with index of 1 i would want it to be 2

site/pages/api/duplicate-consequence.api.ts
```diff
@@ -0,0 +1,73 @@
...
+        );
+
+        if (!consequenceToDuplicate) {
+            throw new Error("No consequence found to duplicate");
+        }
+
+        await upsertConsequence(
+            {
+                ...consequenceToDuplicate,
+                consequenceIndex: disruption.consequences
```
alvak1, line 48: Add disruption.consequences > 0

alvak1 (approved):
Approved

PR #160, opened 2023-07-05
Add display id to disruptions to uniquely identify disruptions

alvak1:
Testing instructions

- Create a new disruption or edit an existing disruption should create a display id. This is visible when you export data in view-all-disruptions page or visible in dashboards page


site/utils/siri.ts
```diff
@@ -127,7 +127,16 @@ export const getPtSituationElementFromDraft = (disruption: Disruption) => {
                           Severity: consequence.disruptionSeverity,
                           Affects: {
                               ...(consequence.consequenceType === "networkWide" ||
-                              consequence.consequenceType === "operatorWide"
+                              consequence.consequenceType === "stops"
+                                  ? {
+                                        Operators: {
+                                            AllOperators: "",
+                                        },
+                                    }
```
alainajaffer, line 135: why is this change here?

alainajaffer:
None of my old disruptions are showing

alvak1:
> None of my old disruptions are showing

Yes thats cause displayId is a mandatory field and disruptions without this field will fail parsing

site/utils/siri.ts
alvak1: Fixed after merging with main

PR #174, opened 2023-07-13
Fix for issue with reverting the datasource to default when sys admin updates an org

alvak1:
The fix resolves the issue of datasource reverting to default value when a change is made to the organisation

### DIFF
--- a/site/pages/sysadmin/org.page.tsx
+++ b/site/pages/sysadmin/org.page.tsx
@@ -13,7 +13,7 @@ import { getOrganisationInfoById } from "../../data/dynamo";
 import { fetchAdminAreaCodes } from "../../data/refDataApi";
 import { PageState } from "../../interfaces";
 import { Organisation, organisationSchema, areaCodeSchema, AreaCodeValuePair } from "../../schemas/organisation.schema";
-import { destroyCookieOnResponseObject, getPageState } from "../../utils/apiUtils";
+import { destroyCookieOnResponseObject } from "../../utils/apiUtils";
 import { getSessionWithOrgDetail } from "../../utils/apiUtils/auth";
 import { getStateUpdater } from "../../utils/formUtils";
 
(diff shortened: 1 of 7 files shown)


site/pages/sysadmin/org.page.tsx
```diff
@@ -160,12 +161,16 @@ export const getServerSideProps = async (
...
-        "ID",
-        !errorCookie && orgId && orgInfo ? orgInfo : undefined,
-    );
+    let pageState: PageState<Partial<Organisation>> = {
+        errors: [],
+        inputs: {},
+    };
+
+    if (errorCookie) {
+        pageState = JSON.parse(errorCookie) as PageState<Partial<Organisation>>;
```
alainajaffer, line 170: could we not use zod to parse this?

site/pages/sysadmin/org.page.tsx
```diff
@@ -117,6 +117,7 @@ const ManageOrgs = (props: PageState<Partial<ManageOrgProps>>): ReactElement =>
 
                 <input type="hidden" name="PK" value={pageState.inputs.PK} />
                 <input type="hidden" name="adminAreaCodes" value={pageState.inputs.adminAreaCodes ?? []} />
+                <input type="hidden" name="mode" value={JSON.stringify(pageState.inputs.mode)} />
```
alainajaffer, line 120: will mode always be defined otherwise this would fail?

site/pages/sysadmin/org.page.tsx
alvak1: As the errorCookie is json string we can use zod to parse it

site/pages/sysadmin/org.page.tsx
alvak1: yes it will always be returned because the org will always have this info in db

alainajaffer (approved):

PR #185, opened 2023-07-21
fix: Remove deleting users restriction

alainajaffer:
Removing the error that gets generated when your a sysadmin deleting a user from another org

### DIFF
--- a/site/pages/api/admin/delete-user.api.ts
+++ b/site/pages/api/admin/delete-user.api.ts
@@ -6,7 +6,8 @@ import {
     SYSADMIN_ADD_USERS_PAGE_PATH,
     USER_MANAGEMENT_PAGE_PATH,
 } from "../../../constants";
-import { deleteUser as deleteCognitoUser } from "../../../data/cognito";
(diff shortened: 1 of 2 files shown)


site/pages/api/admin/delete-user.api.ts
```diff
@@ -27,14 +26,6 @@ const deleteUser = async (req: DeleteUserApiRequest, res: NextApiResponse): Prom
             throw Error("Insufficient values provided to delete a user");
         }
 
-        const userDetails = await getUserDetails(username);
-
-        const formattedUserDetails = user.parse(userDetails);
-
-        if (formattedUserDetails.organisation !== session.orgId) {
```
ljones92, line 34: We should probably keep this check in and only disable it in the case where the user is a sysadmin:

```ts
if (!session.isSystemAdmin && formattedUserDetails.organisation !== session.orgId)
```

alainajaffer: done

ljones92 (approved):

PR #187, opened 2023-07-24
feat: Duplicate disruption - information banner

alainajaffer:
Go to an existing disruption and hit duplicate disruption and the banner should show then click publish or save as draft and navigate back to the duplicate there should be no banner


alvak1 (approved):
Approved

PR #188, opened 2023-07-25
Added delete disruption button to more pages

ljones92:


alainajaffer (approved):
Checked that button is in all files and works on different pages

PR #190, opened 2023-07-25
fix: Polygon tool - Services 

alainajaffer:
a user shouldn't be able to use both the polygon shape and the services drop down at the same time
services/stops should be cleared when a polygon is drawn
when a service is selected from the dropdown and a polygon had already been drawn prior but not selected the selected services stops will be selected not the polygon's stops

### DIFF
--- a/site/components/map/ServicesMap.tsx
+++ b/site/components/map/ServicesMap.tsx
@@ -261,7 +261,7 @@ const Map = ({
 
                 const servicesToAdd = servicesInPolygon
                     .filter((service) =>
-                    service.stops.filter((stop) => markerData.map((marker) => marker.atcoCode).includes(stop))
+                        service.stops.filter((stop) => markerData.map((marker) => marker.atcoCode).includes(stop)),
                     )
                     .map((service) => serviceSchema.parse(service));
 
--- a/site/components/map/ServicesMap.tsx
+++ b/site/components/map/ServicesMap.tsx
@@ -181,6 +181,8 @@ const Map = ({
 
                 if (stopsData) {
                     setMarkerData(stopsData);
+                    setSelectedServices([]);
+                    clearServicesAndStops();
                 } else {
                     setMarkerData([]);
                 }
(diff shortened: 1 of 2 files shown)


site/components/map/ServicesMap.tsx
```diff
@@ -222,120 +224,129 @@ const Map = ({
         setPopupInfo({});
     }, []);
 
-    const selectAllStops = async (evt: SyntheticEvent) => {
-        evt.preventDefault();
+    const addSelectedStopsAndServices = async (includeMarkerData?: boolean) => {
```
DeannaSharma, line 227: If I use the polygon tool first and then I select a service, the polygon tool remains. Is this the intended functionality or should the polygon tool be cleared?

See screen recording below:
https://github.com/Department-for-Transport-Disruptions/create-disruptions-data/assets/83244002/1d0714bf-650e-4c0d-8700-130450357814

site/components/map/ServicesMap.tsx
```diff
@@ -222,120 +224,129 @@ const Map = ({
...
+                              return {
+                                  ...marker,
+                                  serviceIds: services.length > 0 ? services.map((s) => s.id) : undefined,
+                              };
+                          })
+                    : [];
+
+                const servicesToAdd = servicesInPolygon
+                    .filter((service) =>
+                        service.stops.filter((stop) => {
```
DeannaSharma, line 264: because .includes returns a true or false value we don't need the if statement

`service.stops.filter((stop) => markerData.map((marker) => marker.atcoCode).includes(stop))`

site/components/map/ServicesMap.tsx
```diff
@@ -222,120 +224,129 @@ const Map = ({
...
+                          .map((marker) => {
+                              const services = servicesInPolygon.filter((service) =>
+                                  service.stops.includes(marker.atcoCode),
+                              );
+                              return {
+                                  ...marker,
+                                  serviceIds: services.length > 0 ? services.map((s) => s.id) : undefined,
+                              };
+                          })
+                    : [];
```
DeannaSharma, line 260: The servicesMap function is getting pretty significant, it might be best to break it out while we're at this stage 
and introduce some unit tests to as much of the new business logic if possible, if we can break out little sections it'll make it more readable and the tests will give us confidence that this area of the code is working as expected (for future devs as much as for us!)

i.e. breaking out the functionality here into a `getMarkerDataInAService` function and unit testing that separately.

```
const markerDataInAService = includeMarkerData ? getMarkerDataInAService(markerData, servicesStopsInPolygon, servicesInPolygon) : []

function getMarkerDataInAService(markerData, servicesStopsInPolygon, servicesInPolygon) {
    return markerData
        .filter((marker) => servicesStopsInPolygon.includes(marker.atcoCode))
        .map((marker) => {
            const services = servicesInPolygon.filter((service) =>
                service.stops.includes(marker.atcoCode),
            );
            return {
                ...marker,
                serviceIds: services.length > 0 ? services.map((s) => s.id) : undefined,
            };
        })
}
```

DeannaSharma (changes requested):

site/components/map/ServicesMap.tsx
alainajaffer: intended it just shouldn't select stops from the polygon and it should select those in that service

site/components/map/ServicesMap.tsx
alainajaffer: done the refactoring for this - anything I should be specifically testing?

DeannaSharma: Awesome, I would create a unit test for getMarkerDataInAService and test all the possible paths i.e.

```
describe('getMarkerDataInAService', () => {
  it('should return an empty array if markerData is an empty array', () => {})
  it('should return an empty array if no matching atcoCode is found within serviceStopsInPolygon', () => {})
  it('should return an array of markers with undefined serviceIds if a matching atcoCode is found in serviceStopsInPolygon but is not found in serviceInPolygon', () => {})
  it('should return markerData within a given service if atcoCode is found in serviceStopsInPolygon and serviceInPolygon', () => {})
})
```

Hope that makes sense but give me a shout if you want to go through it together!

alainajaffer: done

DeannaSharma (approved):

PR #192, opened 2023-07-26
feat: Filter by rail

alainajaffer:


site/components/map/ServicesMap.tsx
```diff
@@ -176,6 +176,8 @@ const Map = ({
                         ? { stopTypes: ["MET", "PLT"] }
                         : vehicleMode === Modes.ferry || vehicleMode === VehicleMode.ferryService
                         ? { stopTypes: ["FER", "FBT"] }
+                        : vehicleMode === Modes.rail
```
DeannaSharma, line 179: Ticket only mentions adding rail to stop level consequence. Does it need to be included in services level consequences also?

If so these lines of code also need adding to /create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx to allow Rail stations to be selected via the drop down.

site/components/map/StopsMap.tsx
```diff
@@ -116,6 +116,8 @@ const Map = ({
                         ? { stopTypes: ["MET", "PLT"] }
                         : vehicleMode === Modes.ferry || vehicleMode === VehicleMode.ferryService
                         ? { stopTypes: ["FER", "FBT"] }
+                        : vehicleMode === Modes.rail
```
DeannaSharma, line 119: I think these lines of code also need adding to /create-consequence-stops/[disruptionId]/[consequenceIndex].page.tsx to allow Rail stations to be selected via the drop down.

site/components/map/ServicesMap.tsx
alainajaffer: just checked with steven - just stops consequence page

site/components/map/StopsMap.tsx
alainajaffer: done

DeannaSharma (approved):

PR #206, opened 2023-08-08
feat: UI polygon large

alainajaffer:
Draw a polygon on stop and services consequences if too large you will see a message

Links to https://github.com/Department-for-Transport-Disruptions/reference-data-service/pull/44


site/components/map/ServicesMap.tsx
```diff
@@ -484,19 +488,32 @@ const Map = ({
               )} (${service.operatorShortName})`
             : "Line: N/A";
     };
+    useEffect(() => {
```
ljones92, line 491: Since we're already doing this check in the API, rather than import the library here as well and do the same check it might be nicer to just look at the api response for the "Area of polygon must be below" error and handle it that way

alainajaffer: done

site/data/refDataApi.ts
```diff
@@ -42,6 +42,10 @@ export const fetchStops = async (input: FetchStopsInput) => {
         method: "GET",
     });
 
+    if (!res.ok) {
+        return (await res.json()) as { error: string };
```
ljones92, line 46: Another way to do this might be to use some custom errors which can then be caught by whatever's invoking this function, so this could be:

```ts
if (!res.ok) {
    if (body.error.includes("Area of polygon must be below")) {
        throw new LargePolygonError();
    }

    throw new Error(`fetchStops call failed: ${body.error}`);
}
```

where `LargePolygonError` could be defined in `site/errors/index.ts` and would look like:

```ts
export class LargePolygonError extends Error {
    constructor() {
        super();
        Object.setPrototypeOf(this, LargePolygonError.prototype);
    }
}
```

then anywhere calling it can be wrapped in a try/catch:

```ts
try {
    ...
}
catch (e) {
    if (e instanceof LargePolygonError) {
        setWarningMessage("Drawn area too big, draw a smaller area");
    } else {
        setWarningMessage("There was a problem retrieving the stops");
    }
}
```

alainajaffer: done

site/data/refDataApi.ts
```diff
@@ -43,7 +44,16 @@ export const fetchStops = async (input: FetchStopsInput) => {
...
+    if (!res.ok) {
+        const body = (await res.json()) as { error: string };
+        if (body.error.includes("Area of polygon must be below")) {
+            throw new LargePolygonError();
+        }
+
+        throw new Error(`fetchStops call failed: ${body.error}`);
+    }
+
+    const parseResult = z.array(stopSchema).safeParse(await res.json());
```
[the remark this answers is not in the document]
alainajaffer: done

PR #214, opened 2023-08-11
fix: improve selecting stops from polygon functionality when in service consequence

DeannaSharma:
- Fixed bug: when a stop was selected from polygon the services corresponding to that stop was not being populated in the dropdown. 
- Added a warning message to pop up when a stop is selected from the polygon tool that does not have an associated service
- Refactored warning message function


site/components/map/ServicesMap.tsx
```diff
@@ -280,30 +316,21 @@ const Map = ({
...
-                    .filter((service) =>
-                        service.stops.filter((stop) => markerData.map((marker) => marker.atcoCode).includes(stop)),
-                    )
-                    .map((service) => serviceSchema.parse(service));
-
                 setSelectedServices(
                     [
                         ...(selectedServices ?? []),
-                        ...(includeMarkerData
+                        ...(!!markerData
```
alainajaffer, line 333: if markerData is an empty array this will still run do we want a length check too?

DeannaSharma: added length check

alainajaffer (approved):

PR #215, opened 2023-08-11
fix: Map error placement

alainajaffer:
Warning message to be under the select all button

### DIFF
--- a/site/components/map/StopsMap.tsx
+++ b/site/components/map/StopsMap.tsx
@@ -231,9 +231,6 @@ const Map = ({
 
     return mapboxAccessToken ? (
         <>
-            {selected.length === 100 ? (
-                <Warning text={`Stop selection capped at 100, ${selected.length} stops currently selected`} />
-            ) : null}
             {showSelectAllButton ? (
                 <button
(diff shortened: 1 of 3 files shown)


site/components/map/StopsMap.tsx
```diff
@@ -245,6 +241,10 @@ const Map = ({
                     {showSelectAllText ? "Select all stops" : "Unselect all stops"}
                 </button>
             ) : null}
+            {selected.length === 100 ? (
+                <Warning text={`Stop selection capped at 100, ${selected.length} stops currently selected`} />
+            ) : null}
+            {warningMessage ? <Warning text={warningMessage} /> : null}
```
DeannaSharma, line 247: We should probably implement the same warningMessageText logic to stops map. (Sorry I should have done that in my last ticket)

I'd recommend creating a mapUtils files and moving warningMessageText from ServicesMap.tsx to that utils file. Then replicating the warningMessage logic from ServiceMap.tsx in StopsMap.tsx

alainajaffer: done

DeannaSharma (approved):

PR #221, opened 2023-08-21
fix: Remove duplicates in stops list when more than one service

alainajaffer:
go to services consequence select the example in the screenshot using 099 admin code - you should not see duplicate stops


site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx
```diff
@@ -229,7 +229,16 @@ const CreateConsequenceServices = (props: CreateConsequenceServicesProps): React
     useEffect(() => {
         if (selectedService) {
             fetchStops(selectedService.id, pageState.inputs.vehicleMode, dataSource)
-                .then((stops) => setStopOptions(sortStops([...stopOptions, ...stops])))
+                .then((stops) =>
+                    setStopOptions(
+                        sortStops([
+                            ...stopOptions.filter(
```
Comment (DeannaSharma) [inline site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx:235]: Can we move this sortStops and filter logic to fetchStops?

On line 651 we use fetchStops without sorting and filtering which means unordered and duplicate stops will appear if you go back to this page after entering data to edit your changes.

site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx
Comment (DeannaSharma) [inline site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx:1]: On Line 58 remove the export as this function is not used elsewhere in the codebase.

site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx
Comment (DeannaSharma) [reply site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx:235]: Also, it might be a good idea to create a new function in the FormUtils file to do both the sorting and remove duplicate actions, following a similar patter to the filterServices function. That way we have a reusable piece of code that can easily be tested. 

I would use this function on this consequence page, and also use this new function in the ServicesMap file and the StopsMap file wherever we currently sort and filter stops.

site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx
Comment (alainajaffer) [reply site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx:1]: done

site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx
Comment (alainajaffer) [reply site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx]: done

site/components/map/StopsMap.tsx
```diff
@@ -209,7 +209,7 @@ const Map = ({
                         ...state,
                         inputs: {
                             ...state.inputs,
-                            stops: sortStops(
+                            stops: sortAndFilterStops(
                                 [...selected, ...markerData]
```
DeannaSharma, line 213: Remove this .filter as it's doing repeating the same logic as sortAndFilterStops(). To return the first 100 values move the .splice. See below:

```
stops: sortAndFilterStops([...selected, ...markerData]).splice(0, 100),
```

alainajaffer: done

DeannaSharma (approved):

PR #225, opened 2023-08-23
fix: Remove frontend validation

alainajaffer:
- Also solves: Error message appears when you click on calendar icon after clicking type box for start date

Testing instructions:

Check each page there should not be on blur events a client side validation


site/components/form/DateSelector.tsx
```diff
@@ -147,19 +133,7 @@ const DateSelector = <T extends object>({
...
-                                inputId,
-                                inputName,
-                                errors,
-                                disabled,
-                                stateUpdater,
-                                setErrors,
-                                schema,
-                                errorOnBlur,
-                            );
+                            return inputBox(inputRef, inputProps, InputProps, inputId, inputName, errors, disabled);
```
alvak1, line 136: RangeError is thrown when you submit a new disruption without date and time fields.

site/components/form/Select.tsx
```diff
@@ -67,15 +66,8 @@ const Select = <T extends object>({
...
-                        onBlur={
-                            !updateOnChange
-                                ? (e) => handleBlur(e.target.value, inputName, stateUpdater, setErrors, schema)
-                                : undefined
-                        }
                         onChange={
-                            updateOnChange
-                                ? (e) => handleBlur(e.target.value, inputName, stateUpdater, setErrors, schema)
-                                : undefined
+                            updateOnChange ? (e) => handleChange(e.target.value, inputName, stateUpdater) : undefined
```
alvak1, line 70: We could use stateUpdater function instead of handleChange as this function invokes the stateUpdater without any  other changes.

alainajaffer: done

site/components/form/DateSelector.tsx
ljones92: @alvak1 This is actually a bug on main, I've fixed it in my PR

alainajaffer: pulled main in the fix is now in

alvak1 (approved):
Approved

PR #242, opened 2023-09-01
Component alignment updates to create disruption page

alvak1:
Navigate to create new disruptions page and click save without filling one or all of the mandatory date fields


site/components/form/DateSelector.tsx
```diff
@@ -33,7 +33,7 @@ const inputBox = <T extends object>(
     disabled: boolean,
 ) => (
     <div className="govuk-date-input flex flex-row [&_.MuiSvgIcon-root]:fill-govBlue">
-        <div className="govuk-date-input__item govuk-!-margin-right-0">
+        <div className="govuk-date-input__item govuk-!-margin-right-0]">
```
ljones92, line 36: Does the "]" do anything here?

DeannaSharma: Alignment looks very slightly off for the date boxes as they are all different lengths. Perhaps we could add a min width to this input, something like: ```<div className="govuk-date-input__item govuk-!-margin-right-0 min-w-[240px]">```

This would mean increasing the min width for for the divs containing the DateSelector components in [disruptionId].page.tsx - see other comment

I think we should also remove the "]" as it adjusts the alignment in a weird way.

site/pages/create-disruption/[disruptionId].page.tsx
```diff
@@ -322,7 +331,7 @@ const CreateDisruption = (props: DisruptionPageProps): ReactElement => {
                             </Fragment>
                         ))}
                         <div className="flex pb-8">
-                            <div>
+                            <div className={`${dateColumnError ? "min-w-[300px] pr-1" : "pr-1"}`}>
```
Comment (DeannaSharma) [inline site/pages/create-disruption/[disruptionId].page.tsx:334]: As per other comment, bump this up to 320px or something similar if min width is added to the input. Repeat elsewhere in this file where min width is 300px

site/components/form/DateSelector.tsx
alvak1: done

site/pages/create-disruption/[disruptionId].page.tsx
Comment (alvak1) [reply site/pages/create-disruption/[disruptionId].page.tsx:334]: done

site/pages/create-disruption/[disruptionId].page.tsx
Comment (DeannaSharma) [inline site/pages/create-disruption/[disruptionId].page.tsx:1]: Is it possible to align the "Now" links? Just had another look at the ticket and Sami left a comment about this.

Comment (alvak1) [reply site/pages/create-disruption/[disruptionId].page.tsx:1]: done

DeannaSharma (approved):

PR #247, opened 2023-09-05
Fix width of column while sorting

alvak1:
Fixed width of columns in the table displayed in view all disruptions page


site/pages/__snapshots__/view-all-disruptions.test.tsx.snap
```diff
@@ -3492,13 +3492,13 @@ exports[`pages > viewAllDisruptions > should render correctly when there are no
...
             <th
-              className="govuk-table__header align-middle px-2"
+              className="govuk-table__header align-middle px-2 w-[12%]"
               scope="col"
             >
               Severity
             </th>
             <th
-              className="govuk-table__header align-middle px-2"
+              className="govuk-table__header align-middle px-2 w-[12%]"
```
alainajaffer, line 3501: could we use w-2/12 or w-1/12 a built in tailwind class instead ?

alvak1: This is also a built in functionality from tailwind and the math is easier to change width

alainajaffer (approved):

PR #250, opened 2023-09-06
fix: Green spot on map disappears when selected

alainajaffer:
On stops consequence map only

Expected: when you search for a stop and select an option on the menu that spot turns green you can then unselect it and it will go grey and stay on the map - when the stop is grey if you search something else the stop will disappear.


site/components/map/StopsMap.tsx
DeannaSharma, line 1: As discussed, when I unselect a stop it turns grey as expected but I cannot then reselect the stop.

DeannaSharma (approved):

PR #253, opened 2023-09-08
Fixed width of input boxes and divs when errors are passed to create disruption page

alvak1:
Fixed width of input boxes and divs when errors are passed to create disruption page


site/components/form/DateSelector.tsx
DeannaSharma, line 1: The date and time fields are different heights when in an error state, maybe we can fix the heights?

Also the calendar icon and now buttons are positioned far away from the input box - can we confirm with Chris if this is ok or if they should be positioned closer to the input box.

Screenshot 2023-09-12 at 09 24 34

DeannaSharma, line 1: At smaller screen sizes, the input boxes overflow (it starts to when the screen width is around 800px). If we're fixing widths maybe at smaller screen sizes we can align the input fields in a single column to prevent it overflowing.

Screenshot 2023-09-12 at 09 27 24

alvak1: done

DeannaSharma (approved):

DeannaSharma (approved):

PR #254, opened 2023-09-11
feat: sys admin view all users

DeannaSharma:
Added functionality for system admins to be able to view all user types for a given organisation. System admins should be able to remove all user types, but can only resend invites to org admins. 

System admins should be able to view other system admins, but not perform any actions.

### DIFF
--- a/site/pages/admin/user-management.page.tsx
+++ b/site/pages/admin/user-management.page.tsx
@@ -37,7 +37,9 @@ const UserManagement = ({ userList, csrfToken }: UserManagementPageProps): React
                     `${getAccountType(user.group)}`,
                     user.email,
                     user.userStatus === "CONFIRMED" ? "Active" : "Pending invite",
-                    createLink("user-action", index, user.username, user.group, user.userStatus !== "CONFIRMED"),
+                    user.group !== "system-admins"
+                        ? createLink("user-action", index, user.username, user.group, user.userStatus !== "CONFIRMED")
+                        : "",
                 ],
             });
         });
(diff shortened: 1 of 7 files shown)


alvak1:
The acceptance criteria 1 states that the header should be of format “Add {Organisation name} admin”

site/pages/sysadmin/users.page.tsx
```diff
@@ -38,20 +39,23 @@ const AdminUsers = (props: AdminUserProps): ReactElement => {
 
     const getRows = () => {
         const rows: { header?: string | ReactNode; cells: string[] | ReactNode[] }[] = [];
-        props.admins?.forEach((user, index) => {
+        props.users?.forEach((user, index) => {
```
alvak1, line 42: Could the data be sorted according to first name or email? Chris or Anna should be able to confirm

site/pages/admin/user-management.page.tsx
```diff
@@ -78,42 +68,37 @@ const UserManagement = ({ userList, csrfToken }: UserManagementPageProps): React
         userGroup: string,
         showResendInvite?: boolean,
     ) => {
-        {
-            if (userGroup !== "system-admins") {
-                return (
+        return (
+            <>
+                {showResendInvite ? (
```
alvak1, line 73: Could System admin users benefit from having resent invite incase of unregistered status. If the invitation for a system admin user has expired then there should be an option to resend the invite.

In case of system admins who are no longer part of the team - there should be an option to delete them

DeannaSharma: Spoke with Chris yesterday about this, he confirmed that system admins and org admins should only be able to view system admins and perform no actions. 

I think there will be a separate ticket about system admin functionality as ideally we don't want them to be a part of an organisation at all.

site/pages/sysadmin/users.page.tsx
DeannaSharma: refactored and created sortedAndFilteredUsersList to sort users by first name when we perform the filter action

alvak1 (approved):
Approved

PR #261, opened 2023-09-15
fix: Error page found when change link is selected on existing template

alainajaffer:
The issue was that if you go to an existing template and attempt to change something it would error.

The fix is to check whether the template is in edit or create flow and then redirect accordingly.

Test the create flows and the edit flows.

Also fixes Error page found when changing the selected consequence during creating a new template

### DIFF
--- a/site/pages/api/create-consequence-services.api.ts
+++ b/site/pages/api/create-consequence-services.api.ts
@@ -98,7 +98,9 @@ const createConsequenceServices = async (req: NextApiRequest, res: NextApiRespon
         destroyCookieOnResponseObject(COOKIES_CONSEQUENCE_SERVICES_ERRORS, res);
 
         const redirectPath =
-            !isFromTemplate && queryParam && decodeURIComponent(queryParam).includes(DISRUPTION_DETAIL_PAGE_PATH)
+            (!isFromTemplate || template) &&
+            queryParam &&
+            decodeURIComponent(queryParam).includes(DISRUPTION_DETAIL_PAGE_PATH)
                 ? DISRUPTION_DETAIL_PAGE_PATH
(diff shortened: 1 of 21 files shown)


site/pages/api/create-consequence-services.api.ts
DeannaSharma, line 1: I still get an error when I click the "Change" link to change the consequence type.

I think on line 401 href needs a bit of logic to pass ```?template=true``` if we are in template mode. 

This will also need changing for the "Change" link for other consequence types

alainajaffer: done

site/utils/formUtils.ts
```diff
@@ -122,7 +122,5 @@ export const showCancelButton = (queryParams: ParsedUrlQuery) => {
 };
 
 export const returnTemplateOverview = (queryParams: ParsedUrlQuery) => {
-    return (
-        queryParams["return"]?.includes(DISRUPTION_DETAIL_PAGE_PATH) && queryParams["return"]?.includes("template=true")
-    );
+    return queryParams["return"]?.includes(DISRUPTION_DETAIL_PAGE_PATH) && queryParams["template"]?.includes("true");
```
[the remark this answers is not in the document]
DeannaSharma: This is what the code should be on /type-of-consequences page, line 69 for the href:
```
href={
      returnToTemplateOverview
         ? `${queryParams["return"] as string}/${pageState.disruptionId || ""}${isTemplate ? "?template=true" : "" }`
         : `${queryParams["return"] as string}/${pageState.disruptionId || ""}`
}
```

DeannaSharma (approved):

PR #265, opened 2023-09-18
Feature/disruption 392 twitter api

ljones92:


site/pages/create-social-media-post/[disruptionId]/[socialMediaPostIndex].page.tsx
Comment (DeannaSharma) [inline site/pages/create-social-media-post/[disruptionId]/[socialMediaPostIndex].page.tsx:1]: Request from Chris:

Could we have it where if you load the page, then you see the twitter/hootsuite drop down and the message content field. If you select twitter, nothing else happens and you're good to go. If you actually select hootuite, then the extra stuff appears like hootsuite profile, image, time/date etc?

DeannaSharma (dismissed):

DeannaSharma (approved):
🚢 it

PR #278, opened 2023-09-21
feat: Get organisation endpoint

alainajaffer:


packages/get-organisation/index.ts
```diff
@@ -0,0 +1,63 @@
+import { getOrganisationInfoAndStats } from "@create-disruptions-data/shared-ts/utils/dynamo";
```
ljones92, line 1: I think we could have both of these endpoints in the same package folder, maybe something like `organisations-api/get-organisation` and `organisations-api/get-organisations`. Then they can share the same package.json and config files etc.

shared-ts/utils/dynamo.ts
```diff
@@ -219,3 +219,52 @@ export const getAllOrganisationsInfoAndStats = async (logger: Logger): Promise<O
...
     }
 };
+
+export const getOrganisationInfoAndStats = async (
+    orgId: string,
+    logger: Logger,
+): Promise<OrganisationsWithStats[0] | null> => {
+    logger.info(`Getting organisation ${orgId} with stats from DynamoDB table...`);
+    try {
+        const dbDataInfo = await recursiveScan(
```
ljones92, line 229: We should use a `recursiveQuery` here instead of a scan

shared-ts/utils/dynamo.ts
```diff
@@ -219,3 +219,52 @@ export const getAllOrganisationsInfoAndStats = async (logger: Logger): Promise<O
...
+                ExpressionAttributeValues: {
+                    ":info": "INFO",
+                    ":stat": "STAT",
+                    ":orgId": orgId,
+                },
+            },
+            logger,
+        );
+
+        const orgIds = [...new Set(dbDataInfo.map((item) => item.PK))];
```
ljones92, line 242: We don't need to collect the org ids here since there's only the one, we can just have:

```ts
const info = dbDataInfo.find((item) => item.SK === "INFO");
const stats = dbDataInfo.find((item) => item.SK === "STAT");

const parsedOrgWithStats = organisationSchemaWithStats.safeParse({
    ...info,
    stats
});
```

We'd need a new schema for this:
```ts
export const organisationSchemaWithStats = organisationSchema.and(z.object({ stats: statisticSchema }));
```

stacks/services/APIGateway.ts
```diff
@@ -37,6 +37,7 @@ export const createSiriApi = (stack: Stack, siriSXBucket: Bucket, hostedZone: IH
         },
         routes: {
             "GET    /organisations": "packages/get-organisations/index.main",
+            "GET    /organisations/{id}": "packages/get-organisation/index.main",
```
ljones92, line 40: Then these paths will need updating with the new folder structure

alainajaffer: done

packages/get-organisation/index.ts
alainajaffer: done

ljones92 (approved):

PR #282, opened 2023-09-25
feat: update to service consequence functionality

DeannaSharma:
Updated services consequence functionality so that when a stop is selected from the dropdown menu, all services associated with that stop are added to the services dropdown.

### DIFF
--- a/site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx
+++ b/site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx
@@ -220,8 +220,37 @@ const CreateConsequenceServices = (props: CreateConsequenceServicesProps): React
                     includeRoutes: true,
                     dataSource: dataSource,
                 });
+
+                const servicesRoutesForGivenStop = servicesForGivenStop?.map((service) => {
+                    return {
+                        inbound: service.routes.inbound,
+                        outbound: service.routes.outbound,
+                        serviceId: service.id,
+                    };
+                });
+
+                const servicesRoutesForMap = [...searched, ...servicesRoutesForGivenStop].filter(
+                    (value, index, self) =>
+                        index === self.findIndex((service) => service?.serviceId === value?.serviceId),
+                );
+
+                const stopsForServicesRoutes = await Promise.all(
+                    servicesRoutesForMap.map(async (service) => {
+                        if (service) {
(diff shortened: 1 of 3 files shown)


site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx
```diff
@@ -220,11 +220,49 @@ const CreateConsequenceServices = (props: CreateConsequenceServicesProps): React
             });
         } else {
             if (stopToAdd) {
+                const servicesForGivenStop = await fetchServicesByStops({
+                    atcoCodes: [stopToAdd.atcoCode],
+                    includeRoutes: true,
+                    dataSource: dataSource,
+                });
```
Comment (alainajaffer) [inline site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx:227]: When multiple services are connected to a stop, when I add a stop then delete the service that I originally searched for the stop attached gets removed but the other services remain - is that what we expect?
For example:
Searched for 1 - Barnsley Interchange - Staincross Circular (Stagecoach)
Selected 1 - Barnsley Interchange - Staincross Circular (Stagecoach)
Searched Bar Lane/Cloudberry Way adj 370050414
Selected Bar Lane/Cloudberry Way adj 370050414
Delete 1 - Barnsley Interchange - Staincross Circular (Stagecoach)
Bar Lane/Cloudberry Way adj 370050414 gets deleted but the 486 - Darton College - Mapplewell (Globe Holidays) and 488 - Darton Academy - Monk Bretton (Globe Holidays) remain

Comment (DeannaSharma) [reply site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx:227]: Updated dropdown logic so that a stop stays selected until all services that include that stops are removed.

Also updated the logic on the stop dropdown options to ensure only options for selected services are shown. 

(Previously, if you selected a stop that added some additional services the stop dropdown options would be updated to include these additional services. However, if you deleted one of those additional services their stop options would sometimes remain in the stop options dropdown)

Comment (alainajaffer) [reply site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx:227]: The select all button for Barnsley Interchange - Staincross Circular (Stagecoach) does not select all related services to every stop

Comment (DeannaSharma) [reply site/pages/create-consequence-services/[disruptionId]/[consequenceIndex].page.tsx:227]: Chris confirmed that when clicking the select all button we should only go and grab all the stops related to the selected services. We shouldn't select all the related services to all the stops.

alainajaffer (approved):

PR #284, opened 2023-09-26
fix: 500 error when deleting consequence from existing template

DeannaSharma:


site/pages/disruption-detail/[disruptionId].page.tsx
```diff
@@ -331,7 +331,8 @@ const DisruptionDetail = ({
         });
     };
 
-    const nextIndex = getLargestConsequenceIndex(disruption) + 1;
+    const nextIndex =
+        disruption.consequences && disruption.consequences.length > 0 ? getLargestConsequenceIndex(disruption) + 1 : 0;
```
Comment (ljones92) [inline site/pages/disruption-detail/[disruptionId].page.tsx:335]: Is this check needed? It's also done within the `getLargestConsequenceIndex` function

Comment (ljones92) [reply site/pages/disruption-detail/[disruptionId].page.tsx:335]: Also it'll have the side effect of overwriting deleted consequences

Comment (DeannaSharma) [reply site/pages/disruption-detail/[disruptionId].page.tsx:335]: Without this check, If we delete a consequence from a disruption so no consequence exists then create a new consequence the index will be 1 not 0 which I think is incorrect?

I've seen this cause a bug when creating a disruption from a template - can show you on a call.

Comment (ljones92) [reply site/pages/disruption-detail/[disruptionId].page.tsx:335]: Historically that was intentional so that the consequence index would always be unique within a disruption so it wouldn't end up overwriting previous consequences. That might not be an issue anymore with how edits are handled  now though. Happy to jump on a call to see the bug it's causing.

Comment (DeannaSharma) [reply site/pages/disruption-detail/[disruptionId].page.tsx:335]: reverted it back to original code

site/pages/review-disruption/[disruptionId].page.tsx
```diff
@@ -325,14 +325,18 @@ const ReviewDisruption = ({
         });
     };
 
-    const nextIndex = getLargestConsequenceIndex(disruption) + 1;
+    const nextIndex =
+        disruption.consequences && disruption.consequences.length > 0 ? getLargestConsequenceIndex(disruption) + 1 : 0;
```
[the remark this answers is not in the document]
Comment (DeannaSharma) [reply site/pages/review-disruption/[disruptionId].page.tsx:329]: reverted back to original code

ljones92 (approved):

PR #296, opened 2023-10-02
redirect to target page on login

ljones92:


site/middleware.api.ts
```diff
@@ -182,6 +183,10 @@ export async function middleware(request: NextRequest) {
 
             const newResponse = NextResponse.redirect(new URL(LOGIN_PAGE_PATH, request.url));
 
+            const referer = request.headers.get("referer") ?? new URL(DASHBOARD_PAGE_PATH, request.url).toString();
+
+            newResponse.cookies.set(COOKIES_LOGIN_REDIRECT, !request.url.includes("/api/") ? request.url : referer);
+
```
alvak1, line 189: Works with redirection of links that is when you click on a link and the user is redirected to another page. Does not work when cookies are cleared and you click save button in create disruption page

ljones92: What's happening for you in that circumstance? It should take you back to the page in the state before you clicked save since otherwise we'd need to save the values to post to the api

alvak1 (approved):
Approved

PR #324, opened 2023-10-23
stop creation of consequences after reaching max

ljones92:


site/pages/create-consequence-network/[disruptionId]/[consequenceIndex].page.tsx
```diff
@@ -200,7 +200,7 @@ const CreateConsequenceNetwork = (props: CreateConsequenceNetworkProps): ReactEl
                             returnPath={returnPath}
                         />
 
-                        {(props.consequenceIndex || 0) <= 10 && (
+                        {(props.consequenceCount || 0) < MAX_CONSEQUENCES && (
```
Comment (DeannaSharma) [inline site/pages/create-consequence-network/[disruptionId]/[consequenceIndex].page.tsx:203]: Logic needs modifying slightly. When creating a 15th consequence the "add another consequence" button shows because the 15th disruption isn't in the database yet. This means you can go through the flow to create a 16th consequence - although it will eventually fail on the api check

site/pages/disruption-detail/[disruptionId].page.tsx
Comment (DeannaSharma) [inline site/pages/disruption-detail/[disruptionId].page.tsx:1]: Minor bug found for template flow, steps to recreate:
- Go to view all templates
- Click on an existing template that has 15 consequences
- Manually create a 16 consequence by changing the URL to /type-of-consequence/[disruptionId]/16?[queryParams]
- Go through the flow to create a 16th consequence and hit "Save" button
- Disruption not found page is shown - expected behaviour is to see the create consequence page with the error message of "max consequences reached"

DeannaSharma (approved):

PR #326, opened 2023-10-23
feat: edit user (contains operator user changes)

DeannaSharma:


site/components/page-templates/UserDetailPageTemplate.tsx
```diff
@@ -0,0 +1,239 @@
...
+                                            ? !pageState.inputs.operatorNocCodes.find(
+                                                  (selectedOperators) => selectedOperators.id === operatorOption.id,
+                                              )
+                                            : true,
+                                    )}
+                                    handleChange={handleOperatorChange}
+                                    tableData={pageState.inputs.operatorNocCodes}
+                                    getRows={getOperatorRows}
+                                    getOptionValue={(operator: OperatorData) => operator.id.toString()}
+                                    display="NOC Codes"
```
alainajaffer, line 203: In the wireframe it says NOC Code maybe confirm what is expected here?

site/utils/apiUtils/index.ts
```diff
@@ -99,7 +100,7 @@ export const getPageState = <T>(errorCookie: string, schemaObject: z.ZodType<T>,
         return { ...(JSON.parse(errorCookie) as PageState<Partial<T>>), disruptionId: disruptionId || "" };
     }
 
-    if (disruptionId) {
+    if (disruptionId || !!data) {
```
alainajaffer, line 103: should this be &&? what if there is no data do we want to throw an error?

PR #330, opened 2023-10-26
feat: edit user (without operator user changes)

DeannaSharma:


site/utils/apiUtils/index.ts
```diff
@@ -99,7 +99,7 @@ export const getPageState = <T>(errorCookie: string, schemaObject: z.ZodType<T>,
         return { ...(JSON.parse(errorCookie) as PageState<Partial<T>>), disruptionId: disruptionId || "" };
     }
 
-    if (disruptionId) {
+    if (disruptionId || !!data) {
```
DeannaSharma, line 102: Original comment: should this be &&? what if there is no data or disruptionId do we want to throw an error?

I used an or expression because when I use this function for the edit-user page I don't have a disruptionId to pass through. Looking at the usages there is at least one occasion where we pass through a disruptionId without data so I think this logic is ok and we do not need to throw an error.

site/components/form/TextInput.test.tsx
```diff
@@ -51,4 +51,19 @@ describe("TextInput", () => {
...
+    it("should render correctly if input is disabled", () => {
+        const tree = renderer
+            .create(
+                <TextInput<TestInputs>
+                    inputName="field1"
+                    display="Test Field"
+                    stateUpdater={vi.fn()}
+                    widthClass="w-3/4"
+                    maxLength={50}
+                    disabled={true}
```
alainajaffer, line 63: It is signing me out when i change my name - i assumed it would only make sense to logout if you change the role?

DeannaSharma: missed a bit of logic out on the edit-user api - fixed

alainajaffer (approved):

PR #333, opened 2023-10-31
feat: operator user create disruption

DeannaSharma:
**Wait until tickets 526 and 510 are completed before reviewing this code**

Created new operator user feature branch as the scope of work for adding an operator user has changed meaning the old feature branch is no longer valid.

This PR is for operator users to create a disruption - follows same flow as other users however operators can only create operator and service consequences. 

When creating operator consequence they should only see their own NOC codes.
When creating service consequences they should only see services that contain their NOC codes.


alainajaffer:
image

Delete disruption button is showing on create journey

site/pages/type-of-consequence/__snapshots__/type-of-consequence.test.tsx.snap
```diff
@@ -1,5 +1,353 @@
...
+                type="hidden"
+              />
+              <div
+                className="govuk-button-group"
+              >
+                <button
+                  className="govuk-button"
+                  data-module="govuk-button"
+                >
+                  Save and continue
```
alainajaffer, line 152: When i press save and continue after entering all fields it does not navigate to the review page

alainajaffer (approved):

PR #338, opened 2023-11-01
fix: "Edit" link within the user management page does not work

alainajaffer:
After looking at logs:
{"level":"error","message":"Failed to retrieve group for user"}
'no identity-based policy allows the cognito-idp:AdminListGroupsForUser action' fixing this error

### DIFF
--- a/stacks/SiteStack.ts
+++ b/stacks/SiteStack.ts
@@ -146,6 +146,7 @@ export function SiteStack({ stack }: StackContext) {
                     "cognito-idp:AdminDeleteUser",
                     "cognito-idp:AdminCreateUser",
                     "cognito-idp:ListUsersInGroup",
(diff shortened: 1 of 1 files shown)


stacks/SiteStack.ts
```diff
@@ -146,6 +146,7 @@ export function SiteStack({ stack }: StackContext) {
                     "cognito-idp:AdminDeleteUser",
                     "cognito-idp:AdminCreateUser",
                     "cognito-idp:ListUsersInGroup",
+                    "cognito-idp:AdminListGroupsForUser",
```
DeannaSharma, line 149: Think we need to add "cognito-idp:AdminRemoveUserFromGroup" as well as I forgot to add that in.

DeannaSharma (approved):

PR #342, opened 2023-11-03
feat: Account settings operator

alainajaffer:
When the operator user clicks the “My Account” button

Then the operator user should only be presented with the following options:

- Account settings
- Social media
- Sign out

Given an operator user is within the “Account Settings” page

Then they should see their name (needs to be included in design) alongside noc code and fields as per wireframes


site/components/layout/Header.tsx
```diff
@@ -15,6 +15,7 @@ interface HeaderProps {
 const Header = ({ session, csrfToken }: HeaderProps): ReactElement => {
     const { buttonProps, itemProps, isOpen, setIsOpen } = useDropdownMenu(session?.isOrgAdmin ? 4 : 2);
 
+    console.log(session);
```
DeannaSharma, line 18: remove console.log :)

alainajaffer: done

site/schemas/operator.schema.ts
DeannaSharma, line 1: Could we move this operator schema over to the organisations schema file? 

I like what you did with the transform so I implemented it in my add user PR, I added the below code to the organisations schema file:

```
export const operatorOrgSchema = subOrganisationSchema.transform((data) => ({
    orgId: data.PK,
    operatorOrgId: data.SK.replace("OPERATOR#", ""),
    name: data.name,
    nocCodes: data.nocCodes,
}));

export type OperatorOrgSchema = z.infer<typeof operatorOrgSchema>;

export const operatorOrgListSchema = z.array(operatorOrgSchema);
```

alainajaffer: done

site/schemas/session.schema.ts
```diff
@@ -11,19 +11,23 @@ export const sessionSchema = z
...
         "cognito:groups": z.array(z.nativeEnum(UserGroups)).optional(),
     })
     .transform((item) => {
         const isSystemAdmin = item["cognito:groups"]?.includes(UserGroups.systemAdmins) ?? false;
         const isOrgAdmin = item["cognito:groups"]?.includes(UserGroups.orgAdmins) ?? false;
         const isOrgPublisher = item["cognito:groups"]?.includes(UserGroups.orgPublishers) ?? false;
         const isOrgStaff = item["cognito:groups"]?.includes(UserGroups.orgStaff) ?? false;
+        const isOperatorUser = item["cognito:groups"]?.includes(UserGroups.operators) ?? false;
 
         return {
```
DeannaSharma, line 24: Might need some clarification on AC3 for this ticket which is about setting the datasource for operator users. 

At the moment, the datasource will be whatever the LTA org has set - if we follow the ACs then we will need some logic on line 45 to set the datasource for operator users

alainajaffer: done

DeannaSharma (approved):

PR #355, opened 2023-11-16
fix: Creation time

alainajaffer:


shared-ts/disruptionTypes.zod.ts
```diff
@@ -785,6 +785,13 @@ export const consequenceSchema = z.discriminatedUnion("consequenceType", [
 
 export const MAX_CONSEQUENCES = 15;
 
+const historySchema = z.object({
```
DeannaSharma, line 788: If this is identical to the historySchema that is in `disruption.schema.ts` then we should delete the schema in the `disruption.schema.ts` and update its usages to use the schema in the shared-ts folder to avoid duplicating code.

packages/siri-sx-generator/util/siri.ts
```diff
@@ -63,6 +63,11 @@ export const getPtSituationElementFromSiteDisruption = (
     disruption: Disruption & { organisation: { id: string; name: string } },
 ) => {
     const currentTime = getDate().toISOString();
+    const creationTime = disruption.creationTime
```
DeannaSharma, line 66: I think we should tweak the logic to retrieve the creation time via the value `historyItems` rather than sorting the dates. 

If we consider a disruption to be created when it is published then we should grab the datetime of the historyItem which says "Disruption created and published"

If a disruption requires approval before being published, then at the moment the logic will bring back when a disruption was sent for approval not when it was published.

DeannaSharma: Also as part of this I would create a new enum in `enums.ts` called historyItem and list all the possible options of history item texts and update the strings we pass through so that we know exactly what the value will be.

alainajaffer: done

shared-ts/disruptionTypes.zod.ts
alainajaffer: done

PR #367, opened 2023-12-01
feature: roadworks overview page

DeannaSharma:


site/schemas/roadwork.schema.ts
```diff
@@ -0,0 +1,88 @@
...
+export type WorkStatus = z.infer<typeof workStatus>;
+
+export const worksLocationType = [
+    "Footway",
+    "Carriageway",
+    "Verge",
+    "Footpath",
+    "Parking place",
+    "Bus stop or stand",
+    "Cycle hire docking station",
```
alainajaffer, line 23: should we be using literals here too?

DeannaSharma: This was a copied over from ref data service but after taking another look we're not actually using the worksLocation type so I'll remove it since its not needed.

alainajaffer:
Clarify what happens if there are no roadworks

site/pages/view-all-roadworks.page.tsx
```diff
@@ -0,0 +1,115 @@
...
+        };
+    });
+};
+
+const ViewAllRoadworks = ({ liveRoadworks }: ViewAllRoadworksProps) => {
+    return (
+        <BaseLayout title={title} description={description}>
+            <h1 className="govuk-heading-xl">Roadworks in your area</h1>
+
+            {liveRoadworks.length === 0 && <Warning text={"There are no current roadworks in your area"} />}
```
alainajaffer, line 60:
```suggestion
            {liveRoadworks.length === 0 && <Warning text="There are no current roadworks in your area" />}
```

site/data/refDataApi.ts
```diff
@@ -273,3 +274,52 @@
...
+interface FetchRoadworkByIdInput {
+    permitReferenceNumber: string;
+}
+
+export const fetchRoadworkById = async (input: FetchRoadworkByIdInput) => {
+    const searchApiUrl = `${API_BASE_URL}/roadworks/${input.permitReferenceNumber}`;
+
+    const res = await fetch(searchApiUrl, {
+        method: "GET",
+    });
```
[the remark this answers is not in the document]
DeannaSharma: A user input (Permit reference number) has to be provided in order to query the roadworks endpoint to see if it exists. Permit reference numbers from street manager do not have a strictly defined format hence the regex expression to simply check whether the permitReferenceNumber includes only numbers, letters, _ , . or -

PR #368, opened 2023-12-01
feature: roadwork detail page

DeannaSharma:


site/utils/index.ts
```diff
@@ -214,3 +214,9 @@ export const getStopsForRoutes = async (
         )
     ).flat();
 };
+
+export const toTitleCase = (text: string) => {
+    return text.replace(/([^\W_]+[^\s-]*) */g, function (txt) {
```
alainajaffer, line 219: could we use an arrow function here?

site/utils/index.ts
```diff
@@ -214,3 +214,9 @@ export const getStopsForRoutes = async (
...
+        return txt.charAt(0).toUpperCase() + txt.substring(1).toLowerCase();
```
alainajaffer, line 220: could this be defensive what if txt is empty?

site/utils/index.test.ts
```diff
@@ -133,3 +133,15 @@ describe("filterDisruptionsForOperatorUser", () => {
...
     });
 });
+
+describe("toTitleCase", () => {
+    it.each([
+        ["KING STREET", "King Street"],
+        ["king street", "King Street"],
+        ["HYPHENATED-STREET", "Hyphenated-Street"],
+        ["hyphenated-street", "Hyphenated-Street"],
+        ["RanDoM cAsE", "Random Case"],
```
alainajaffer, line 143: A case for empty string

DeannaSharma: added

site/utils/index.ts
DeannaSharma: if empty string is passed an empty string is returned

alainajaffer (approved):

PR #378, opened 2023-12-14
feat: Reporting

alainajaffer:


site/utils/exportUtils.ts
```diff
@@ -73,5 +73,25 @@ export const getExportSchema = (): Schema<ExportDisruptionData> => {
...
         },
+        {
+            column: "Planned/Unplanned",
+            type: String,
+            value: (objectData: ExportDisruptionData) => objectData.disruptionType,
+        },
+        {
+            column: "Date added",
+            type: String,
+            value: (objectData: ExportDisruptionData) => objectData.creationTime,
```
DeannaSharma, line 84: Old disruption don't have a creation time so when I export the data a lot of disruptions have N/A under "date added"

In the SIRI generator we have a function called `getDisruptionCreationTime` where if there is not a creation time we use the disruption history to get when the disruption was created. Can we reuse this function to populate the date added column?

alainajaffer: done

site/pages/api/get-all-disruptions/__snapshots__/get-all-disruptions.test.ts.snap
```diff
@@ -9,6 +9,7 @@ exports[`getAllDisruptions > formatSortedDisruptions > correctly formats disrupt
   "displayId": "8fg3ha",
   "disruptionReason": "accident",
   "disruptionType": "planned",
+  "history": undefined,
```
DeannaSharma, line 12: do we expect this to be undefined? we might need to update the mock data the test is using to include a mock history item.

alainajaffer: done

DeannaSharma (approved):

PR #379, opened 2023-12-15
fix: Error page when entering letters into time and date fields on create disruption page

alainajaffer:


shared-ts/disruptionTypes.zod.ts
DeannaSharma, line 1: It might be worth changing the default Zod error for disruptionStartTime and publicationStartTime to be "Invalid start time" and "Invalid publication start time".

Currently when I enter letters in the start time fields the error message is "Enter a start time"

site/pages/api/create-disruption.test.ts
```diff
@@ -367,6 +367,338 @@ describe("create-disruption API", () => {
...
+                "",
+                "weekly",
+                getFutureDateAsString(40),
+            ],
+        };
+        const { req, res } = getMockRequestAndResponse({ body: disruptionData, mockWriteHeadFn: writeHeadMock });
+
+        await createDisruption(req, res);
+
+        const errors: ErrorInfo[] = [{ errorMessage: "Enter publication start time", id: "publishStartTime" }];
```
DeannaSharma, line 399: Similar to comment above, this error message doesn't make complete sense for an invalid inout. If we change/ add a custom error message we should update its usages within the tests.

alainajaffer: done

shared-ts/disruptionTypes.zod.ts
alainajaffer: done

DeannaSharma (approved):

PR #387, opened 2023-12-19
chore: Upgrade to node 20

alainajaffer:


package.json
```diff
@@ -22,21 +22,22 @@
...
+        "constructs": "^10.3.0",
         "eslint": "^8.34.0",
         "eslint-config-prettier": "^8.6.0",
         "eslint-import-resolver-typescript": "^3.5.3",
         "eslint-plugin-import": "^2.27.5",
         "eslint-plugin-prettier": "^4.2.1",
         "husky": "^8.0.0",
         "prettier": "^2.8.4",
-        "sst": "2.30.2",
+        "sst": "^2.38.7",
```
ljones92, line 40: Better to keep the versions of sst, cdk and constructs pinned so it doesn't end up making infrastructure changes without us knowing beforehand

stacks/SiriAPIStack.ts
```diff
@@ -38,7 +38,7 @@ export const SiriAPIStack = ({ stack }: StackContext) => {
         defaults: {
             function: {
                 timeout: 20,
-                runtime: "nodejs18.x",
```
ljones92, line 41: We'll also need to add the runtime to the `NextjsSite` construct in the site stack or it'll keep using node 18 which is still the default

stacks/SiriGeneratorStack.ts
```diff
@@ -46,7 +46,7 @@ export const SiriGeneratorStack = ({ stack }: StackContext) => {
         handler: "packages/siri-sx-generator/index.main",
         timeout: 60,
         memorySize: 1536,
-        runtime: "nodejs18.x",
+        runtime: "nodejs20.x",
```
ljones92, line 49: This will need adding to the stats generator function as well

package.json
alainajaffer: done

stacks/SiriAPIStack.ts
alainajaffer: done

stacks/SiriGeneratorStack.ts
alainajaffer: done

package.json
```diff
@@ -22,21 +22,22 @@
...
+        "@types/node": "^20.10.5",
         "@typescript-eslint/eslint-plugin": "^6.8.0",
         "@typescript-eslint/parser": "^6.8.0",
         "@vitest/coverage-c8": "^0.29.2",
-        "aws-cdk": "2.95.1",
-        "aws-cdk-lib": "2.95.1",
-        "constructs": "10.2.69",
+        "aws-cdk": "2.110.1",
+        "aws-cdk-lib": "2.110.1",
+        "constructs": "^10.3.0",
```
ljones92, line 32: Keep this one pinned as well

alainajaffer: done

ljones92 (approved):

PR #390, opened 2023-12-20
feat: Services dropdown

alainajaffer:


site/components/form/SearchSelect.tsx
```diff
@@ -32,6 +32,8 @@ interface SearchSelectProps<T> {
     width?: string;
     onFocus?: () => void;
     onBlur?: () => void;
+    closeMenuOnSelect?: boolean;
+    backspaceRemovesValue?: boolean;
```
DeannaSharma, line 36: What's the expected behaviour for `backspaceRemovesValue`? 

I don't think this has added any different functionality to what was there before but I just want to double check in case I've misunderstood how it's meant to work.

site/components/form/SearchSelect.tsx
DeannaSharma, line 1: I've noticed an issue where when you search for a value and the little cross icon appears to the left of the search bar if you then click the cross button it does not clear what you've searched for.

alainajaffer: fixed

site/components/form/SearchSelect.tsx
alainajaffer: removed

DeannaSharma (approved):

PR #395, opened 2024-01-05
async load disruptions on dashboard

ljones92:


site/pages/dashboard.test.tsx
DeannaSharma, line 1: Might be worth adding unit test for the `formatDisruptions()` function since we've removed the tests from this file.

DeannaSharma (approved):

PR #402, opened 2024-01-11
load disruptions from s3 for dashboard

ljones92:


site/data/s3.ts
```diff
@@ -31,7 +31,7 @@ export const putItem = async (
     }
 };
 
-export const getItem = async (bucket: string, key: string, originalFilename: string): Promise<string> => {
+export const getS3SignedUrl = async (bucket: string, key: string, originalFilename: string): Promise<string> => {
     logger.info("", {
         context: "data.s3",
         message: "getting item from s3",
```
DeannaSharma, line 37: Should we change the logger message and error message to "getting signed url from s3" - just thinking that `getS3SignedUrl` and `getObject` has the same error message which may make debugging annoying?

site/pages/api/get-all-disruptions/[organisationId].api.ts
```diff
@@ -24,179 +17,41 @@ export interface GetDisruptionsApiRequest extends NextApiRequest {
...
-        if (!worstSeverity) {
-            worstSeverity = severity;
-        } else if (severityScoringMap[worstSeverity] < severityScoringMap[severity]) {
-            worstSeverity = severity;
-        }
-    });
-
-    return worstSeverity;
-};
+export const getDashboardDisruptions = async (orgId: string, isTemplate: boolean) => {
```
Comment (DeannaSharma) [inline site/pages/api/get-all-disruptions/[organisationId].api.ts:20]: Maybe we could change the name of this function as we don't just use it on the dashboard page? 

I would have used `getAllDisruptions` but that's already the parent function's name! Maybe something like `fetchDisruptionsData`?

DeannaSharma (approved):

PR #403, opened 2024-01-16
feat: Roadworks map

alainajaffer:


alainajaffer:
Note: Having an issue where a typescript issue is showing on the github actions but not on my local.

site/components/map/RoadWorksMap.tsx
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        return null;
+    };
+
+    const onClick = useCallback((event: MapLayerMouseEvent) => {
+        const feature = event.features?.[0];
+
+        if (feature?.layer.id === "clusters") {
+            if (feature && feature.properties) {
```
alvak1, line 122: Could these two if conditions be combined into one?

alainajaffer: done

site/components/map/RoadWorksMap.tsx
```diff
@@ -0,0 +1,206 @@
...
+                    if (!map.hasImage("cone") && image) map.addImage("cone", image, { sdf: false });
+                });
+            }
+        }
+
+        return null;
+    };
+
+    const onClick = useCallback((event: MapLayerMouseEvent) => {
+        const feature = event.features?.[0];
```
alvak1, line 119: Could we map feature variable to type `MapboxGeoJSONFeature` which might allow us to remove some type assertions below

alainajaffer: this change causes errors below

site/pages/view-all-roadworks.page.tsx
```diff
@@ -17,6 +23,60 @@ export interface ViewAllRoadworksProps {
...
+        const worksLocationCoordinates: GeoJSONFeature = {
+            type: "Feature",
+            geometry: {
+                type: "Point",
+                coordinates: [0, 0],
+            },
+        };
+
+        const geometry = parse(item.worksLocationCoordinates || "") || worksLocationCoordinates.geometry;
+        worksLocationCoordinates.geometry = geometry;
```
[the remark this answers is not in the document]
alainajaffer: Just overriding the defaults with the correct value here. I am parsing the string from the api into geojson there are different types at this point polygon, linestring and point

PR #412, opened 2024-02-09
feat:  street manager notification email settings

DeannaSharma:
Note: Might be worth waiting until ticket DISRUPTION-597 (notification emails for new roadworks) is completed before merging this PR.

### DIFF
--- a/site/pages/account-settings.page.tsx
+++ b/site/pages/account-settings.page.tsx
@@ -115,39 +115,14 @@ const AccountSettings = ({
         );
     };
 
-    const updateDisruptionApprovalEmailPreferences = async (emailPreference: boolean) => {
-        setDisruptionApprovalEmailPreference(emailPreference);
-        const url = new URL("/api/admin/update-email-preference", window.location.origin);
-        csrfToken ? url.searchParams.append("_csrf", csrfToken) : null;
-        const res = await fetch(url.toString(), {
-            method: "POST",
-            headers: csrfToken
-                ? {
-                      "Content-Type": "application/json",
-                      "X-CSRF-TOKEN": csrfToken,
-                  }
-                : { "Content-Type": "application/json" },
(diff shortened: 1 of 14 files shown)


site/pages/api/admin/update-email-preference.api.ts
```diff
@@ -16,8 +16,6 @@ const updateEmailPreference = async (req: NextApiRequest, res: NextApiResponse)
 
         if (!session) {
             throw new Error("No session found");
-        } else if (!session.isOrgAdmin) {
-            throw new Error("Invalid user accessing the page");
```
alainajaffer, line 20: Just wondering if the file path should be changes if we aren't checking if they are not an admin anymore. It is still api/admin...

alainajaffer: Perhaps we still want to check groups based on what email preference we are updating

alainajaffer, line 20: Emily said For any preferences e.g. email notifications or data sources, we only present these choices to organisation admin accounts currently - so I think we should still have this check

site/pages/account-settings.page.tsx
```diff
@@ -222,34 +239,38 @@ const AccountSettings = ({
                               ]
                     }
                 />
-                {sessionWithOrg.isOrgAdmin ? (
```
alainajaffer, line 225: I think this check should still apply

site/pages/api/admin/update-email-preference.api.ts
DeannaSharma: Checked with Emily and she's happy to leave code as is so we can set email preferences for street manager for org admins, publishers and staff.

alainajaffer (dismissed):

alainajaffer (approved):

PR #427, opened 2024-03-13
fix: Duplicate operator list

alainajaffer:


site/components/ViewAllContents.tsx
```diff
@@ -702,7 +702,11 @@ const ViewAllContents = ({
 
         setCombinedServicesList(combinedServices);
 
-        setOperatorsList(operators);
+        const uniqueOperatorArray = operators.filter(
+            (obj, index, self) => index === self.findIndex((t) => t.id === obj.id),
```
DeannaSharma, line 706:
```suggestion
            (value, index, self) => index === self.findIndex((operator) => operator.id === value.id),
```

Could be nice to updating naming to follow the pattern we've established elsewhere in the codebase.

Also, we do a similar operation to remove duplicate operators on the create-operator-consequence page but in a longer form way. I think it could be could to create a separate function in our formUtils folder called `removeDuplicateOperators` and update our create-consequence page to use this function to ensure consistency in how we remove duplicates.

alainajaffer: done

DeannaSharma (approved):

PR #439, opened 2024-04-16
London Underground

alainajaffer:


site/schemas/organisation.schema.ts
```diff
@@ -16,6 +17,7 @@ export const defaultModes: ModeType = {
     tram: Datasource.tnds,
     ferryService: Datasource.tnds,
     rail: Datasource.tnds,
+    underground: Datasource.tnds,
 };
 
 export const organisationSchema = z.object({
```
DeannaSharma, line 23: On line 27 we can remove `.default(defaultModes)`  and the const `defaultModes` since we are setting defaults individually in `modeSchema` now.

site/testData/mockData.ts
```diff
@@ -877,6 +877,25 @@ export const mockSessionWithOrgDetail: SessionWithOrgDetail = {
     mode: defaultModes,
     isOperatorUser: false,
     operatorOrgId: null,
+    showUnderground: false,
+};
+
+export const mockSessionWithOrgDetailTFL: SessionWithOrgDetail = {
```
DeannaSharma, line 883:
```suggestion
export const mockSessionWithOrgDetailAndShowUndergroundTrue: SessionWithOrgDetail = {
```

I would probably rename and remove the reference to TfL - although we know it is for TfL we haven't referenced a specific org anywhere else in the code base :)

site/pages/view-all-disruptions.page.tsx
```diff
@@ -15,6 +15,7 @@ const ViewAllDisruptions = ({
     filterStatus,
     enableLoadingSpinnerOnPageLoad = true,
     orgId,
+    showUnderground,
```
DeannaSharma, line 18:
```suggestion
    showUnderground = false,
```

We should default this value in case session is not returned.

site/components/ViewAllContents.tsx
```diff
@@ -977,7 +979,9 @@ const ViewAllContents = ({
                                 defaultDisplay="Select a mode"
                                 selectValues={[
                                     { display: "Any", value: "any" },
-                                    ...VEHICLE_MODES.sort((a, b) => a.display.localeCompare(b.display)),
+                                    ...VEHICLE_MODES.filter((v) =>
+                                        showUnderground ? true : v.value !== VehicleMode.underground,
```
DeannaSharma, line 983: We're doing this same filter expression in a lot of places, if possible it might be nice to bring this out into a separate utils function and reference that function throughout the codebase. Maybe call it something like `filterVehicleModes`

alainajaffer: done

site/testData/mockData.ts
alainajaffer: done

site/schemas/organisation.schema.ts
alainajaffer: done

site/utils/index.ts
```diff
@@ -225,3 +226,15 @@ export const removeDuplicates = <T, K extends keyof T>(arrayToRemoveDuplicates:
...
+    VEHICLE_MODES.filter((v) => (showUnderground ? true : v.value !== VehicleMode.underground));
+
+export const filterStopList = (stops: Stop[], vehicleMode: VehicleMode | Modes, showUnderground?: boolean) =>
+    stops.filter((stop) =>
+        showUnderground && vehicleMode === VehicleMode.underground
+            ? stop.commonName.toLowerCase().includes("underground")
+            : showUnderground && vehicleMode === VehicleMode.tram
+            ? stop.commonName.toLowerCase().includes("tram")
+            : true,
+    );
```
DeannaSharma, line 240:
```suggestion
export const filterStopList = (stops: Stop[], vehicleMode: VehicleMode | Modes, showUnderground?: boolean) =>
    stops.filter((stop) => {
        if (showUnderground && vehicleMode === VehicleMode.underground) {
            return stop.commonName.toLowerCase().includes("underground");
        }
        if (showUnderground && vehicleMode === VehicleMode.tram) {
            return stop.commonName.toLowerCase().includes("tram");
        } else return true;
    });
```

I think in this case if statements are a bit more readable than a nested ternary

site/pages/create-consequence-stops/[disruptionId]/[consequenceIndex].page.tsx
```diff
@@ -102,8 +101,16 @@ const CreateConsequenceStops = (props: CreateConsequenceStopsProps): ReactElemen
                             : { stopTypes: ["undefined"] }),
                     });
 
-                    if (stopsData) {
-                        setStopOptions(stopsData);
+                    const filteredStopList = stopsData.filter((stop) =>
```
Comment (DeannaSharma) [inline site/pages/create-consequence-stops/[disruptionId]/[consequenceIndex].page.tsx:104]: We should use the `filterStopList` function here too

DeannaSharma (approved):

PR #440, opened 2024-04-18
Increase limit for next door messages

alainajaffer:


site/pages/create-social-media-post/[disruptionId]/[socialMediaPostIndex].page.tsx
Comment (DeannaSharma) [inline site/pages/create-social-media-post/[disruptionId]/[socialMediaPostIndex].page.tsx:1]: Happy with the code, all works as expected. Just thinking, since we have differing logic depending on account type, it would be good to add some test cases to the `create-social-media-post.api.ts` test file. It might be nice to add two tests to make sure we throw an error if we go above 800 characters for nextdoor posts and 280 characters for all other accounts.

Comment (alainajaffer) [reply site/pages/create-social-media-post/[disruptionId]/[socialMediaPostIndex].page.tsx:1]: done

DeannaSharma (approved):

PR #447, opened 2024-05-08
feat: Published line name

alainajaffer:


packages/siri-sx-generator/util/siri.ts
```diff
@@ -197,6 +197,9 @@ export const getPtSituationElementFromSiteDisruption = (
                                                         OperatorName: service.operatorShortName,
                                                     },
                                                     LineRef: service.lineName.replace(/\s+/g, "_"),
+                                                    ...(!["preprod", "prod"].includes(stage)
```
DeannaSharma, line 200: Similar to test last PR, can we create a new site constant called something like `PUBLISHED_LINE_NAME_FEATURE_FLAG` and use that here so we can find it easier when we want to release this to prod.

shared-ts/siriTypes.zod.ts
```diff
@@ -83,6 +83,7 @@ export const networksSchema = z.object({
                 z.object({
                     AffectedOperator: affectedOperatorSchema,
                     LineRef: z.string(),
+                    PublishedLineName: z.string().optional(),
```
DeannaSharma, line 86: I don't think this should be optional as it is a required field now as part of the SIRI updates

```suggestion
                    PublishedLineName: z.string(),
```

packages/siri-sx-generator/util/siri.ts
DeannaSharma, line 1: I think we should modify our tests as currently are tests do not account for disruptions with service consequences so we can't test this functionality. I've checked the generated SIRI file in S3 and it looks good but since we'll be playing around with this functionality in future some additional test could be beneficial so we know if we introduce any regressions.

To do this I would create a new const in the testData file called `consequenceInfoServiceTest` with details for a service consequence and add this to dbResponse variable. This should allow us to test that in normal cases PublishedLineName is being generated.

Then I would create a new test case for checking when we are in prod or preprod we do not generate a PublishedLineName.

shared-ts/siriTypes.zod.ts
alainajaffer: as discussed

packages/siri-sx-generator/util/siri.ts
alainajaffer: done

packages/siri-sx-generator/util/siri.ts
alainajaffer: done

packages/siri-sx-generator/__snapshots__/index.test.ts.snap
```diff
@@ -344,6 +394,31 @@ exports[`SIRI-SX Generator > correctly generates SIRI-SX XML without Version or
...
+                                <Networks>
+                                    <AffectedNetwork>
+                                        <VehicleMode>rail</VehicleMode>
+                                        <AffectedLine>
+                                            <AffectedOperator>
+                                                <OperatorRef>NOC</OperatorRef>
+                                                <OperatorName>Test</OperatorName>
+                                            </AffectedOperator>
+                                            <LineRef>Line</LineRef>
+                                            <PublishedLineName>Line</PublishedLineName>
```
DeannaSharma, line 410: PublishedLineName is being generated here when we don't expect it. might need checking whether the feature flag is being set correctly.

alainajaffer: fixed

DeannaSharma (approved):

PR #449, opened 2024-05-13
feat: AffectedPlace

alainajaffer:


packages/siri-sx-importer/utils/importerSiriTypes.zod.ts
```diff
@@ -103,6 +103,17 @@ export const networksSchema = z.object({
     }),
 });
 
+export const placesSchema = z.object({
```
DeannaSharma, line 106: Instead of defining this in the importer, you could add the shared-ts as a dependency and then import it from there.

To do this, add the below to the importer's dependcies in the package.json:
`"@create-disruptions-data/shared-ts": "file:../../shared-ts"`

packages/siri-sx-generator/util/siri.ts
```diff
@@ -60,12 +61,14 @@ const getPeriod = (period: Validity): Period => ({
         : {}),
 });
 
-export const getPtSituationElementFromSiteDisruption = (
+export const getPtSituationElementFromSiteDisruption = async (
     disruption: Disruption & { organisation: { id: string; name: string } },
-) => {
+): Promise<PtSituationElement> => {
+    const adminAreas = await fetchAdminAreas();
```
DeannaSharma, line 67: I think in this case it would be better to make the `fetchAdminAreas` API call the next level up and pass `adminAreas` through to the getPtSituationsElementFromSiteDisruption function. The reason for this, is so we only make one API call instead of one API per disruption (could slow the generator down in prod)

packages/siri-sx-generator/index.ts
```diff
@@ -50,12 +50,14 @@ const enrichDisruptionsWithOrgInfo = async (disruptions: Disruption[], orgTableN
...
-const convertJsonToSiri = (
+const convertJsonToSiri = async (
     disruptions: Awaited<ReturnType<typeof enrichDisruptionsWithOrgInfo>>,
     currentTime: string,
     responseMessageIdentifier: string,
 ) => {
-    const ptSituationElements = disruptions.map((disruption) => getPtSituationElementFromSiteDisruption(disruption));
+    const ptSituationElements = await Promise.all(
+        disruptions.map(async (disruption) => await getPtSituationElementFromSiteDisruption(disruption)),
+    );
```
DeannaSharma, line 60: See comment below, it would be better to make the API call here so we only make it once instead of an API call per disruption.

```suggestion
    const adminAreas = await fetchAdminAreas();
    const ptSituationElements = disruptions.map((disruption) => getPtSituationElementFromSiteDisruption(disruption, adminAreas));
```

alainajaffer: done

packages/siri-sx-generator/util/siri.ts
alainajaffer: done

packages/siri-sx-importer/utils/importerSiriTypes.zod.ts
alainajaffer: done

packages/siri-sx-generator/util/index.ts
```diff
@@ -164,3 +165,29 @@ export const includeDisruption = (disruption: Disruption, currentDatetime: Dayjs
...
+    administrativeAreaCode: z.string(),
+    name: z.string(),
+    shortName: z.string(),
+});
+
+const API_BASE_URL = "https://api.test.ref-data.dft-create-data.com/v1";
+
+export type AdminArea = z.infer<typeof adminAreaSchema>;
+
+export const fetchAdminAreas = async () => {
```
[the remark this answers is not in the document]
alainajaffer: done

packages/siri-sx-generator/util/index.ts
```diff
@@ -164,3 +165,29 @@ export const includeDisruption = (disruption: Disruption, currentDatetime: Dayjs
...
     return true;
 };
+
+const adminAreaSchema = z.object({
+    administrativeAreaCode: z.string(),
+    name: z.string(),
+    shortName: z.string(),
+});
+
+const API_BASE_URL = "https://api.test.ref-data.dft-create-data.com/v1";
```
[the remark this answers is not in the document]
alainajaffer: done

PR #454, opened 2024-05-28
fix: Operator account created without relevant restrictions

alainajaffer:
Use the terminal to create an operator account and attempt to logjn as an admin and edit that operator account - an error should be produces


DeannaSharma:
Not for this PR - but our cli-helper command to create an user isn't fit for purpose for creating operator users. I'll add a ticket to the backlog to update the create-user script. Low priority but worth doing at some point :)

DeannaSharma (approved):

PR #461, opened 2024-06-24
Hootsuite profile dropdown text change

alainajaffer:


site/schemas/social-media-accounts.schema.ts
```diff
@@ -1,6 +1,11 @@
 import { z } from "zod";
 
-export const hootsuiteProfileSchema = z.object({ type: z.string(), id: z.string(), socialNetworkId: z.string() });
+export const hootsuiteProfileSchema = z.object({
+    type: z.string(),
+    id: z.string(),
+    socialNetworkId: z.string(),
+    socialNetworkUsername: z.string().or(z.null()),
```
DeannaSharma, line 7:
```suggestion
    socialNetworkUsername: z.string().nullable(),
```

DeannaSharma:
Might be worth adding a comment on the ticket to explain that social media posts created in the past will still use the old ID when you view them on the /review-disruption and /view-all-social media pages and that for testing new posts should be created.

site/schemas/social-media-accounts.schema.ts
alainajaffer: done

DeannaSharma (approved):

PR #467, opened 2024-07-02
fix: Keyboard focus lost when pop-up window appears

alainajaffer:


alainajaffer:
Decided to use a library for this

PR #470, opened 2024-07-04
feat: Stats generator journeys

alainajaffer:


packages/siri-sx-stats-generator/index.ts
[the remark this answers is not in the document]
alainajaffer: done

DeannaSharma, line 1: We should update the `statisticSchema` in `organisationTypes.ts` so that our endpoint brings back journey info. I think it's fine for this not to be behind a feature flag as if we default to zero it won't actually affect the BODS frontend until we make a change on the BODS site to show that data.

```
journeysAffected: z.coerce.number().default(0),
journeysConsequencesCount: z.coerce.number().default(0),
```

packages/siri-sx-stats-generator/utils/statGenerators.ts
```diff
@@ -46,6 +51,18 @@ export const generateConsequenceStats = (key: string, disruption: Disruption) =>
...
+                ...(cancelFeatureFlag
+                    ? {
+                          journeysConsequencesCount:
+                              consequence.consequenceType === "journeys"
+                                  ? acc[key].journeysConsequencesCount + 1
+                                  : acc[key].journeysConsequencesCount,
+                          journeysAffected:
+                              consequence.consequenceType === "journeys"
+                                  ? consequence.journeys.length
+                                  : acc[key].journeysAffected,
```
DeannaSharma, line 63:
```suggestion
                          journeysAffected:
                              consequence.consequenceType === "journeys"
                                  ? acc[key].journeysAffected + consequence.journeys.length
                                  : acc[key].journeysAffected,
```

I think there's a bug with the stats generator and it's not been calculating stopsAffected and servicesAffected correctly. The calculation to get journeys affected should be as above, we should modify the services and stops calculation to follow this logic too.

packages/siri-sx-stats-generator/index.ts
alainajaffer: done

DeannaSharma (approved):

PR #487, opened 2024-08-20
feat: Alerting

alainajaffer:


stacks/SiriGeneratorStack.ts
```diff
@@ -127,6 +140,16 @@ export const SiriGeneratorStack = ({ stack }: StackContext) => {
 
     validatorBucket.addEventNotification(EventType.OBJECT_CREATED, new LambdaDestination(siriValidator));
 
+    siriValidator
```
DeannaSharma, line 143: Can we check if this metric is needed because in the `MonitoringStack.ts` file there already is a `SiriValidationAlarm` that I think is performing the same function

alainajaffer: removed

DeannaSharma (approved):

PR #488, opened 2024-08-27
feat: Create disruption Playwright test

alainajaffer:


playwright.config.ts
```diff
@@ -0,0 +1,79 @@
...
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
+// import dotenv from 'dotenv';
+// dotenv.config({ path: path.resolve(__dirname, '.env') });
+
+/**
+ * See https://playwright.dev/docs/test-configuration.
+ */
+export default defineConfig({
```
nbetts, line 13: It looks like some of this can be removed and/or simplified, perhaps it can be setup more like the config in bods-integrated-data?

playwright/publisher.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    fillConsequenceOperatorPage,
+    fillConsequenceServicePage,
+    fillConsequenceStopsPage,
+    fillCreateDisruptionPage,
+    fillTypeOfConsequencePage,
+} from "./utils";
+
+test.use({ storageState: "playwright/.auth/publisher.json" });
+
+test("create network disruption", async ({ page }: { page: Page }) => {
```
nbetts, line 16:
```suggestion
test("create network disruption", async ({ page }) => {
```
The `test` func is defined by the Playwright library so we don't need to explicitly declare types on the inner func's params. This can can apply to the tests further down as well.

playwright/utils.ts
```diff
@@ -0,0 +1,291 @@
+import { expect, Page, ElementHandle, Locator } from "@playwright/test";
+
+export const checkIfOnDashboard = async (page: Page) => {
+    await page.waitForTimeout(1000);
```
nbetts, line 4: What is the reason for adding manual timeouts (and in other places in this PR)? If the website is slow, perhaps setting a higher global timeout is better

playwright.config.ts
alainajaffer: I think this setup needs to be different as this is for UI testing but I will remove the comments

playwright/utils.ts
alainajaffer: Sometimes the tests glitch in some areas for example when switching between pages or waiting for dropdown options to load so I just added it in the relevant places

nbetts: `waitForTimeout` is discouraged: https://playwright.dev/docs/api/class-page#page-wait-for-timeout

There are better options such as waiting on network events or using more precise selectors. You can also increase the timeout on individual selectors when you know it is going to be a slow event: https://playwright.dev/docs/test-timeouts#introduction

playwright/publisher.spec.ts
alainajaffer: fixed

playwright/utils.ts
alainajaffer: It is causing issues for me to switch to the other approach - let's discuss this

nbetts (approved):

playwright/utils.ts
alainajaffer: Added to tech debt if there is time as an improvement

PR #494, opened 2024-09-06
fix: Add coaches to operators page

alainajaffer:


site/utils/index.ts
```diff
@@ -238,13 +238,27 @@ export const removeDuplicates = <T, K extends keyof T>(arrayToRemoveDuplicates:
         (value, index, self) => index === self.findIndex((item) => item[key] === value[key]),
     );
 
+export const removeDuplicatesBasedOnMode = <T extends Operator, K extends keyof T>(
```
ljones92, line 241: We could do with some tests around this function

alainajaffer: done

ljones92 (approved):

PR #518, opened 2024-12-13
fix: update pdf exporter to use jsPDF

DeannaSharma:


site/pages/_document.page.tsx
```diff
@@ -15,7 +15,7 @@ const generateCsp = (): { csp: string; nonce: string } => {
     const csp = `
         default-src 'self';
         script-src 'self' 'nonce-${nonce}' 'strict-dynamic' ${production ? "" : "'unsafe-eval' 'unsafe-inline'"};
-        connect-src 'self' ${API_BASE_URL}/ https://*.tiles.mapbox.com https://api.mapbox.com https://events.mapbox.com;
+        connect-src 'self' ${API_BASE_URL}/ https://*.tiles.mapbox.com https://api.mapbox.com https://events.mapbox.com data:;
```
nbetts, line 18: can this be scoped down to `data:application/octet-stream`? If not, just `data/application`

DeannaSharma: Ah, I think I'll try removing it entirely now that I'm using jsPDF - i'll look to add it in only if its needed

nbetts (approved):

PR #520, opened 2024-12-16
fix: edits pending approval not displaying on view all disruptions page

DeannaSharma:


shared-ts/db/migrations/1734427558_publish_status_index.mjs
```diff
@@ -0,0 +1,15 @@
+import { Kysely } from "kysely";
+
+/**
+ *
+ * @param {Kysely} db
+ */
+export async function up(db) {
```
ljones92, line 7: Worth putting a rollback step in as well

ljones92 (approved):

PR #541, opened 2025-02-20
refactor(table-renamer): Migrating table-renamer from reference-data-service

ksrk:
Move the table renamer functionality from reference-data-service repo to create-disruptions-data repo
- Lambda function added for table renamer
- Stack added for table renamer


stacks/TableRenamerStack.ts
```diff
@@ -0,0 +1,50 @@
+import { SubnetType } from "aws-cdk-lib/aws-ec2";
+import { Schedule } from "aws-cdk-lib/aws-events";
+import { PolicyStatement } from "aws-cdk-lib/aws-iam";
+import { Cron, Function, StackContext, use } from "sst/constructs";
+
+import { RdsStack } from "./RdsStack";
+import { VpcStack } from "./VpcStack";
+
+export function TableRenamerStack({ stack }: StackContext) {
```
ljones92, line 9: Rather than have this in its own stack, I would define it in the step function stack along with the other lambdas that are used in it.

It will then need adding onto the step function to run at the very end.

stacks/TableRenamerStack.ts
```diff
@@ -0,0 +1,50 @@
...
+        permissions: [
+            new PolicyStatement({
+                actions: ["ssm:PutParameter", "ssm:GetParameter"],
+                resources: ["*"],
+            }),
+        ],
+        enableLiveDev: false,
+    });
+
+    new Cron(stack, "ref-data-service-table-renamer-cron", {
```
ljones92, line 41: We can get rid of the cron since it will run as part of the step function instead

packages/table-renamer/index.ts
```diff
@@ -0,0 +1,162 @@
...
+            await checkReferenceDataImportHasCompleted("operator_public_data", dbClient);
+            await checkReferenceDataImportHasCompleted("operators", dbClient);
+            await checkReferenceDataImportHasCompleted("stops", dbClient);
+            await checkReferenceDataImportHasCompleted("services", dbClient);
+            await checkReferenceDataImportHasCompleted("localities", dbClient);
+            await checkReferenceDataImportHasCompleted("nptg_admin_areas", dbClient);
+
+            await deleteAndRenameTables(dbClient);
+        } else {
+            await putTableRenamerDisableParameter(stage, "false", logger);
```
ljones92, line 53: We should be able to get rid of this now since the error handling will be performed by the step function instead

shared-ts/utils/ssm.ts
```diff
@@ -105,3 +107,20 @@ export const getParametersByPath = async (
         throw error;
     }
 };
+
+export const putTableRenamerDisableParameter = async (stage: string, value: "true" | "false", logger: Logger) => {
```
ljones92, line 111: And we can get rid of this entirely

ljones92 (approved):

ljones92 (approved):

PR #556, opened 2025-05-09
fix: Insert correct timestamps for disruptions validity and publish datetimes

DeannaSharma:


shared-ts/db/migrations/1746785299_timestamps.mjs
```diff
@@ -0,0 +1,45 @@
+import { Kysely } from "kysely";
+
+/**
+ *
+ * @param {Kysely} db
+ */
+export async function up(db) {
+    await db.schema
+        .alterTable("disruptions")
+        .alterColumn("validity_start_timestamp", (col) => col.setDataType("timestamptz"))
```
ljones92, line 10: I think if we do:

```sql
ALTER TABLE disruptions ALTER COLUMN validity_start_timestamp TYPE timestamptz using validity_start_timestamp at time zone 'Europe/London'
```

then it will handle automatically converting the bst dates to utc. Not sure if kysely will support the syntax for this so it may need to be some raw sql.

ljones92 (approved):

PR #584, opened 2025-08-11
chore: Disable ref data migration changes for prod site and functions

DeannaSharma:
Disable the prod site and functions (e.g. SIRI-SX generator and Roadworks Notifier) from using the new disruptions ref data API url. 

This is to unblock releases of other features whilst testing of the new disruptions ref data migration occurs on lower environments.


alvak1 (approved):
Approved

PR #624, opened 2026-05-08
fix: DBODS-1293 - add variables for github actions for new naptan buc…

HarrisonDuffield:
…ket vars

### DIFF
--- a/.github/workflows/manual-deploy.yml
+++ b/.github/workflows/manual-deploy.yml
@@ -74,5 +74,9 @@ jobs:
           MAP_BOX_ACCESS_TOKEN: ${{ secrets.MAP_BOX_ACCESS_TOKEN }}
           SUPPORT_EMAIL: ${{ secrets.SUPPORT_EMAIL }}
           SUPPORT_PHONE: ${{ secrets.SUPPORT_PHONE }}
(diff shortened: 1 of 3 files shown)


.github/workflows/manual-deploy.yml
```diff
@@ -74,5 +74,9 @@ jobs:
           MAP_BOX_ACCESS_TOKEN: ${{ secrets.MAP_BOX_ACCESS_TOKEN }}
           SUPPORT_EMAIL: ${{ secrets.SUPPORT_EMAIL }}
           SUPPORT_PHONE: ${{ secrets.SUPPORT_PHONE }}
+          NAPTAN_BUCKET_NAME: ${{ vars.NAPTAN_BUCKET_NAME }}
```
DeannaSharma, line 77: Just wanted to check have these vars been set at the org level already? I don't have access to check and I can't see them set at the repo level!

HarrisonDuffield: Not set yet - need to catch some of Chris' time today. 

Once the code is in though then it's just a case of set those and redeploy - I hope

Christopher-MW (approved):

DeannaSharma (approved):

PR #632, opened 2026-06-15
fix: DBODS-1927 Amend NPTG schema and amend step function errors

DeannaSharma:
- Amend NPTG zod schema so that DistrictRef is optional
- Amend step function lambda functions to throw errors
- Disable docker builds when running python lambdas locally to fix local dev setup issue with zscaler

### DIFF
--- a/packages/bods-txc-retriever/index.ts
+++ b/packages/bods-txc-retriever/index.ts
@@ -69,7 +69,7 @@ export const main: Handler = async (event, context) => {
         await getBodsDataAndUploadToS3(bodsCoachUrl, txcZippedBucketName, txcBucketName);
     } catch (e) {
         if (e instanceof Error) {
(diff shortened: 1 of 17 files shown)


packages/bods-txc-retriever/index.ts
```diff
@@ -69,7 +69,7 @@ export const main: Handler = async (event, context) => {
         await getBodsDataAndUploadToS3(bodsCoachUrl, txcZippedBucketName, txcBucketName);
     } catch (e) {
         if (e instanceof Error) {
-            logger.error(e);
+            logger.error("There was a problem retrieving bods TXC Data", e);
```
ljones92: The error object should go first in these logger statements

HarrisonDuffield (approved):

